PR #7, opened 2025-10-20
Fixed macOS casing because make.py expects macOS instead of macos

QuasiKili:


ThomasFarstrike:
Perfection. Instant merge!

PR #11, opened 2025-11-19
fixed bug where clicking in swipe areas was ignored

QuasiKili:
Added a short movement detection for back_swipe and top_swipe because otherwise clicks in these areas where ignored. Since I (Quasi Kili) couldn't type the Q on the new mpos_keyboard on an esp32 touchscreen this was an urgent issue. You can test the before and after by clicking on the leftmost part of the q key in the wifi app or the upper bar in the connect 4 app. Sending pressed AND clicked AND released events to the object below might not be the cleanest solution, but the simplest and most universal.

### DIFF
--- a/internal_filesystem/lib/mpos/ui/gesture_navigation.py
+++ b/internal_filesystem/lib/mpos/ui/gesture_navigation.py
@@ -7,16 +7,17 @@ from .display import get_display_width, get_display_height
 downbutton = None
 backbutton = None
 down_start_x = 0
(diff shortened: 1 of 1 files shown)


ThomasFarstrike:
Perhaps this finally opens up the possibility to increase the size of those swipe areas a bit, so that swipes are more easily detected. Currently you have to swipe pretty slowly, to avoid "starting" the swipe outside of the swipe area...

PR #18, opened 2026-01-07
fix macOS build issue

QuasiKili:
script used sed -i '...' without a backup extension, which is incorrect syntax for the BSD sed on macOS, causing a crash:


QuasiKili:
Thanks!

PR #20, opened 2026-01-15
Allow multiple Espressif virtualenv - just use latest

ant9000:
Flashing script fails when multiple Espressif Python virtualenvs are installed. This quick fix just uses the latest.


ThomasFarstrike:
This is great! I guess the old one would fail if there were more than one, so this is a nice improvement!

PR #21, opened 2026-01-20
Update display.py

QuasiKili:
Logo at boot

IMPORTANT: change the url to a correct location 
and put this logo there

https://github.com/QuasiKili/MPOS-logo/blob/main/all_formats/MicroPythonOS-logo-white-long-w240.png


ThomasFarstrike:
Beautiful! I also like the fallback to the text-based MicroPythonOS. I might set a bigger font size on it just to match closer to the logo, although it should never happen anyway :-)

PR #22, opened 2026-01-20
Update run_desktop.sh to fix settings being ignored

QuasiKili:
Script would not be able to start in dark mode, config.json was backed up but never used. config.json is not backed up anymore, change if this is needed still. We overwrite the auto_start_app if we launch it with a specific app.


ThomasFarstrike:
Nice, thank you!

PR #35, opened 2026-02-11
Support ODROID-GO

jedie:
Support for Hardkernel ESP32 device: ODROID-Go (The old one from 2018)

* https://github.com/hardkernel/ODROID-GO/
* https://wiki.odroid.com/odroid_go/odroid_go

What worked:

* Display
* Buttons
* Crossbar
* Wifi
* Battery
* blue LED

TODO:

* Speaker

The blue LED is "coupled" with the button/crossbar press.

### DIFF
--- a/internal_filesystem/lib/mpos/board/odroid_go.py
+++ b/internal_filesystem/lib/mpos/board/odroid_go.py
@@ -1,0 +1,253 @@
+print("odroid_go.py initialization")
+
+# Hardware initialization for Hardkernel ODROID-Go
+# https://github.com/hardkernel/ODROID-GO/
+# https://wiki.odroid.com/odroid_go/odroid_go
+
+import time
+
+import ili9341
+import lcd_bus
+import lvgl as lv
+import machine
+import mpos.ui
+from machine import ADC, Pin
+from micropython import const
+from mpos import InputManager
+
+# Display settings:
+SPI_HOST = const(1)
+SPI_FREQ = const(40000000)
+
+LCD_SCLK = const(18)
+LCD_MOSI = const(23)
+LCD_DC = const(21)
+LCD_CS = const(5)
+LCD_BL = const(32)
+LCD_RST = const(33)
+LCD_TYPE = const(2)  # ILI9341 type 2
+
+TFT_VER_RES = const(320)
+TFT_HOR_RES = const(240)
+
+
+# Button settings:
+BUTTON_MENU = const(13)
+BUTTON_VOLUME = const(0)
+BUTTON_SELECT = const(27)
+BUTTON_START = const(39)
+
+BUTTON_B = const(33)
+BUTTON_A = const(32)
+
+# The crossbar pin numbers:
+CROSSBAR_X = const(34)
+CROSSBAR_Y = const(35)
+
+
+# Misc settings:
+LED_BLUE = const(2)
+BATTERY_PIN = const(36)
+BATTERY_RESISTANCE_NUM = const(2)
+SPEAKER_ENABLE_PIN = const(25)
+SPEAKER_PIN = const(26)
+
+
+print("odroid_go.py turn on blue LED")
+blue_led = machine.Pin(LED_BLUE, machine.Pin.OUT)
+blue_led.on()
+
+
+print("odroid_go.py machine.SPI.Bus() initialization")
+try:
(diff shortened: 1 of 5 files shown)


.gitignore
jedie: @ThomasFarstrike I use PyCharm. So there is the `.idea` directory to exclude. But exclude every variant of these are boring. So i use this idea: Exclude all dot stuff and include the needed one.

Should i create a separate PR only for this change?

internal_filesystem/lib/mpos/main.py
```diff
@@ -57,6 +57,10 @@ def detect_board():
     if sys.platform == "linux" or sys.platform == "darwin": # linux and macOS
         return "linux"
     elif sys.platform == "esp32":
+
+        # TODO: How to detect ODROID-GO ?
```
jedie: Gemini proposal:

```
import machine
import sys

def is_odroid_go():
    # 1. Check if we are even on an ESP32
    if sys.platform != 'esp32':
        return False
    
    try:
        # 2. Check for the specific WROVER flavor (ODROID-Go uses WROVER)
        if 'WROVER' not in sys.implementation._machine:
            return False
        
        # 3. Test a unique pin configuration
        # ODROID-Go has a Blue LED on Pin 2
        led = machine.Pin(2, machine.Pin.OUT)
        
        # 4. Check for the Speaker Pin (GPIO 26) or Game Buttons
        # ODROID-Go buttons are mapped to specific GPIOs (e.g., Menu is 13)
        menu_button = machine.Pin(13, machine.Pin.IN, machine.Pin.PULL_UP)
        
        return True
    except:
        return False

if is_odroid_go():
    print("Hardware detected: Hardkernel ODROID-Go")
```

internal_filesystem/main.py
jedie: This changes can also be move to a separate PR.

internal_filesystem/lib/mpos/board/odroid_go.py
```diff
@@ -0,0 +1,216 @@
...
+BUTTON_A = const(32)
+BUTTON_B = const(33)
+
+# The crossbar settings:
+BUTTON_JOY_X = const(34)
+BUTTON_JOY_Y = const(35)
+
+BUTTON_JOY_MIN_ADC = const(200)
+BUTTON_JOY_MIN_ADC_HIGH = const(2500)
+BUTTON_JOY_MIN_ADC_LOW = const(1200)
```
jedie: The device has a crossbar. Would be really nice to use this for navigation! Currently i just use the Buttons, because it's more complicated to resolve the crossbars, see: https://github.com/jedie/PyOdroidGo/blob/master/odroidgo/crossbar.py

internal_filesystem/lib/mpos/board/odroid_go.py
```diff
@@ -0,0 +1,216 @@
...
+    if button_b.value() == 0:
+        current_key = lv.KEY.PREV
+    elif button_select.value() == 0:
+        current_key = lv.KEY.ENTER
+    elif button_a.value() == 0:
+        current_key = lv.KEY.NEXT
+    elif button_menu.value() == 0:
+        current_key = lv.KEY.ESC
+
+    # Key repeat logic
```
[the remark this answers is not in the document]
jedie: I removed it. Now: Just no auto repeating is possible. But it is so much less code ;)

internal_filesystem/lib/mpos/board/odroid_go.py
```diff
@@ -0,0 +1,217 @@
...
+    if crossbar_pressed := crossbar.poll():
+        key_pressed = crossbar_pressed
+
+    elif button_menu.value() == 0:
+        key_pressed = lv.KEY.BACKSPACE
+    elif button_volume.value() == 0:
+        print("Volume button pressed -> reset")
+        machine.reset()
+    elif button_select.value() == 0:
+        key_pressed = lv.KEY.ESC
```
jedie: It will be pressed, but has currently no effect. e.g.: it's not possible to go back from a started app.

jedie: fixed by calling mpos.ui.back_screen()

ThomasFarstrike:
It's merged, and I had to move around drivers around to make a unified build, but I don't have the hardware to test and make sure I properly adjusted the imports etc.

Would you mind trying the latest esp32 build from https://github.com/MicroPythonOS/MicroPythonOS/actions/ ?

PR #36, opened 2026-02-12
Use a dot-whitelist

jedie:
I use PyCharm. So there is the .idea directory to exclude. But exclude every variant of these are boring. So i use this idea: Exclude all dot stuff and include the needed one.


ThomasFarstrike:
Awesome, you are now an official contributor! :-D

PR #38, opened 2026-02-13
Add simple calendar application

pavelmachek:
Very simple application, but should be suitable for further development. It can add reminders but not remove them for now. I don't believe I got the icon handling right, but I'll need more help with that (icon is in the tree). Probably should be squashed while merging.


ThomasFarstrike:
Thank you! Looks fine to me, ready to merge!

ThomasFarstrike:
I excluded it from scripts/bundle_apps.sh so that it doesn't appear in the AppStore for now because, as you said, it needs further development.

Note that, when I click a date, I get like a weird form with a text field that's always scrolling out of reach, see screenshot.

You can check other apps, like the Wifi app, on how they do it. Or grep for "textarea" in the internal_filesystem/ folder.

Also, I would recommend creating a new "Activity" (like a screen) when the user clicks a date, instead of modifying the current screen. This way, the back swipe action also goes back to the previous Activity, and it keeps each Activity small, modular and readable. See existing apps :-)

calendar

PR #45, opened 2026-02-16
imu: Allow access to iio on Linux

pavelmachek:
Linux phones have IMU units. Allow applications to use them same way they would access IMU on esp32 based devices. Lightly tested on PinePhone.
    
Same applications useful on esp32 (compass, step counter) are likely to be useful on Linux phones, too.


ThomasFarstrike:
You're right, this should be supported! I will try to review it as soon as possible!

PR #96, opened 2026-03-28
Espnow chat app

jedie:


ThomasFarstrike:
Is this ready for merge? I noticed the "ESPNowChat" app is already in the AppStore so... if this fixes something important, we should merge it :-)

jedie:
Sorry the redactor is not ready, yet.

The important part is to be able to switch the WiFi cannel. Because devices are only connected in the same channel.

PR #117, opened 2026-04-08
bugfix "unphone" target in build_mpos.sh

jedie:
fix build error:
```
Error: app partition is too small for binary micropython.bin size 0x3cb940:
- Part 'factory' 0/0 @ 0x10000 size 0x3b9000 (overflow 0x12940)
ninja: build stopped: subcommand failed.
```

by change `partition_size="3900000"` to `partition_size="4194304"` (the default value)

I created a build and flash it: It works fine.


ThomasFarstrike:
Awesome, thank you!

PR #120, opened 2026-04-18
Fix AppearanceManager.set_light_mode / set_primary_color using non-existent prefs.set_string

bitcoin3us:
## Summary
\`AppearanceManager.set_light_mode(is_light, prefs)\` and \`set_primary_color(color, prefs)\` both call \`prefs.set_string(key, value)\`, a method that does not exist on \`SharedPreferences\`. Writes must go through \`prefs.edit().put_string(key, value).commit()\` (see \`lib/mpos/config.py\`).

## Runtime behaviour today

| Call | Result |
|---|---|
| \`set_light_mode(True, prefs)\` | \`AttributeError: 'SharedPreferences' object has no attribute 'set_string'\` |
| \`set_light_mode(True, None)\` | Sets only the in-memory \`_is_light_mode\` flag; LVGL theme is not reinitialised |
| \`set_primary_color(0x123456, prefs)\` | Same \`AttributeError\` |
| \`set_primary_color(0x123456, None)\` | Sets \`_primary_color\` but does not reinitialise the theme, so the colour isn't applied |

So the documented public API is effectively a no-op or raises; it has never worked since \`AppearanceManager\` was added.

## Why it wasn't caught

No code inside \`internal_filesystem/\` calls these setters. The built-in settings app (\`builtin/apps/com.micropythonos.settings/assets/settings.py\`) uses the \`\"ui\": \"radiobuttons\"\` settings framework for persistence and then calls \`AppearanceManager.init(self.prefs)\` from a \`changed_callback\`. Third-party apps following the docstring examples (e.g. a displaywallet I was working on) hit the bug.

## The fix
- Swap \`prefs.set_string(key, value)\` for \`prefs.edit().put_string(key, value).commit()\` in both methods.
- Merge the two \`if prefs:\` branches in \`set_light_mode\` so \`cls.init(prefs)\` is called **after** the save succeeds (previously the second branch was unreachable in the \`prefs\` path because the first always raised).
- Call \`cls.init(prefs)\` from \`set_primary_color\` too — previously it only mutated the in-memory class attr and never reinitialised the LVGL theme, so visually nothing changed even if the save had worked.

Diff is ~13 lines.

## Before / after (behavioural)

**Before:**
\`\`\`python
AppearanceManager.set_light_mode(False, prefs)
# -> AttributeError: 'SharedPreferences' object has no attribute 'set_string'
\`\`\`

**After:**
\`\`\`python
AppearanceManager.set_light_mode(False, prefs)
# -> prefs['theme_light_dark'] = 'dark' on disk
# -> LVGL theme reinitialised in dark mode
# -> [AppearanceManager] Initialized: light_mode=False, ...
\`\`\`

## Verification
- Confirmed the same fixed code pattern works on a real device by using it directly in a third-party app before landing this PR.
- No callers inside MicroPythonOS need changes (none were using the broken methods).

## Related
Third-party impact surfaced during https://github.com/LightningPiggy/LightningPiggyApp/pull/27 (theme toggle in Lightning Piggy), which currently works around the bug by calling \`edit().put_string().commit()\` directly. After this PR, that workaround can be simplified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Frameworks:
 - Add new IRManager framework
 - Add new LoRaManager framework
 - Add mpos.ui.change_task_handler() for improving IR timing accuracy
+- AppearanceManager: fix set_light_mode() and set_primary_color() — they called a non-existent `prefs.set_string()` and raised AttributeError for every third-party caller; writes now go through `edit().put_string().commit()` and the LVGL theme is reinitialised when the colour changes
 
 OS:
 - LilyGo T-Watch S3 Plus: add support for IR Remote app
--- a/internal_filesystem/lib/mpos/ui/appearance_manager.py
+++ b/internal_filesystem/lib/mpos/ui/appearance_manager.py
@@ -143,16 +143,17 @@ class AppearanceManager:
             AppearanceManager.set_light_mode(False)  # Switch to dark mode
         """
         cls._is_light_mode = is_light
-        
-        # Save to preferences if provided
+
+        # Save to preferences if provided, then reinitialise LVGL theme.
+        # SharedPreferences doesn't have a set_string() — writes go through
+        # edit().put_string().commit().
         if prefs:
             theme_str = "light" if is_light else "dark"
-            prefs.set_string("theme_light_dark", theme_str)
-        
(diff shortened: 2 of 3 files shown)


ThomasFarstrike:
Nice catch!

PR #122, opened 2026-04-19
Implement support for Freenove's ESP32-S3 display board

### LINKED ISSUE #119: Please add support for ESP32-S3 2.8" LCD Display
Hello. I have an ESP32-S3 (N16R8) board with capacitive touch as shown below, can you add support for it? I want to try running MicroPythonOS.

Image

Rohansi:
(note: code was written by Copilot but it is working on my device and I can fix whatever you'd like personally)

This PR adds support for [Freenove's ESP32-S3 2.8" display with touch board (FNK0104)](https://store.freenove.com/products/fnk0104). Documentation and sample code can be found here: https://github.com/Freenove/Freenove_ESP32_S3_Display/

It seems to actually just be a generic board that Freenove slapped their branding on. You can find what appears to be the same board listed as ES3C28P on Aliexpress. Freenove's github repo has docs for a board named ES3C28P too.

The board was incorrectly identified as `fri3d_2026` before making any changes, which is why the check for this board is before `fri3d`. I don't know why the board was misidentified though.

The speakers were making a high pitched (but quiet) sound constantly so there are changes in here which disables the DAC when no audio is playing. This should reduce power consumption too. It is not implemented for other boards.

I've personally tested:
* Display
* Backlight brightness
* The boot button (maps to back)
* Touchscreen (single touch - is there anything that needs multitouch?)
* Audio playback
* Audio recording
* LED
* Battery level (it's not super accurate but that may be because we need to use `read_uv()` so the result is calibrated)
* SD card

Should fix https://github.com/MicroPythonOS/MicroPythonOS/issues/119

### DIFF
--- a/internal_filesystem/lib/drivers/codec/es8311.py
+++ b/internal_filesystem/lib/drivers/codec/es8311.py
@@ -46,6 +46,7 @@ _REG13_SYS      = const(0x13)  # system: enable HP output driver
 _REG14_MIC      = const(0x14)  # microphone: DMIC select, analog PGA gain
 _REG17_ADC_VOL  = const(0x17)  # ADC volume / gain
 _REG1C_ADC_EQ   = const(0x1C)  # ADC equalizer bypass + DC-offset cancel
+_REG31_DAC_MUTE = const(0x31)  # DAC soft-mute control (bits[6:5] = 11 → muted)
 _REG32_DAC_VOL  = const(0x32)  # DAC output volume (0x00=muted, 0xFF=max)
 _REG37_DAC_EQ   = const(0x37)  # DAC equalizer / ramp-rate control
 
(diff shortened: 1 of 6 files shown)


Rohansi:
Battery level is wrong. Working on a fix

Rohansi:
Tested everything and it seems good now. Battery level is not very accurate but I don't think there's much that can be done about that without refactoring so that `read_uv()` is used. That would give factory calibrated results that are more accurate.

jedie:
Is this a CYD clone?

Rohansi:
Yes, it's a CYD clone

Rohansi:
Any idea when you'll be able to review this @ThomasFarstrike?

lmoiseichuk:
Probably, this info will be useful: board specs https://www.lcdwiki.com/2.8inch_ESP32-S3_Display
As per https://hosyond.com

ThomasFarstrike:
Sorry for taking so long. This is actually a very nice work! Awesome on the audio driver!

I will take the liberty of adding you to the MAINTAINERS.md file - unless you complain - to keep kind of a record of who added which device, giving credit, and also suggesting that you try out new releases to make sure nothing accidentally broke your board support :-)

PR #125, opened 2026-04-20
SettingActivity: keep exactly one radio option selected

bitcoin3us:
## Summary
- Radio-button groups now preserve the convention that **exactly one option is always selected** once the user has made a choice.
- Clicking the currently-selected option re-checks it instead of deselecting. Changing the pick still works the normal way — click a different option.

## Background
`SettingActivity.radio_event_handler` treated a click on the active option as a toggle-off, setting `active_radio_index = -1` and saving an empty string. That's fine for standalone checkboxes, but wrong for radio groups — downstream code that expected a non-empty value (e.g. Lightning Piggy's `wallet_type` check driving which wallet implementation to construct) would see `""` and fall through to the welcome screen, as if the user had wiped their config.

Users hit this simply by tapping the current selection on the way out of a settings screen — no visible warning, and the saved state was unrecoverable without re-entering all the dependent fields.

## Change
In the `if not checked:` branch of `radio_event_handler`: when the target index matches `active_radio_index`, re-add `STATE.CHECKED` and return. Unrelated paths (switching to a different option, first-time selection from the initial unselected state) are unchanged.

## Test plan
- [x] Lightning Piggy (v0.4.0 branch) on waveshare_esp32_s3_touch_lcd_2: open Settings → Wallet → Wallet Type, tap the currently-selected option — radio stays selected, Save preserves `wallet_type`.
- [x] Tapping the *other* option still switches correctly.
- [x] Any app using `radiobuttons` UI confirms the first-time-selection path still works (no regression for a fresh group with nothing selected yet).


bitcoin3us:
Rebased onto current `main` — conflict was only in `CHANGELOG.md` (other in-flight PRs trimmed their bullets to a shorter final form since this PR was opened). Resolved by keeping main's trimmed versions + inserting the new `SettingActivity:` line in the same position.

Re the earlier failed Linux x64 CI run: the failure was `test_graphical_foreground_app.py` crashing with exit 134 (SIGABRT), which halted `unittest.sh` before the new `test_graphical_setting_activity_radio.py` in this PR would have been reached. That crash appears unrelated to this PR — other platforms (MacOS Intel/arm64, Linux arm64, Linux special-boards) all passed on the same commit, and the crashed test doesn't touch radio buttons or SettingActivity. Just ran `./tests/unittest.sh tests/test_graphical_setting_activity_radio.py` locally on macOS and the three new tests still pass:

```
test_click_active_option_keeps_it_selected ... ok
test_click_different_option_switches_selection ... ok
test_first_selection_from_empty_state ... ok
Ran 3 tests — OK
```

If the fresh Linux x64 run still fails on `test_graphical_foreground_app`, happy to look into that pre-existing flake as a separate PR.

bitcoin3us:
### Checklist close-out — first-time-selection regression guard

The outstanding item:

> Any app using `radiobuttons` UI confirms the first-time-selection path still works (no regression for a fresh group with nothing selected yet).

is directly covered by the unit test shipped in the second commit of this PR — `tests/test_graphical_setting_activity_radio.py::test_first_selection_from_empty_state`:

```python
def test_first_selection_from_empty_state(self):
    """Before any choice is made, active_radio_index = -1. Clicking any
    option selects it normally."""
    # Nothing checked yet; user clicks cb1.
    self.cb1.add_state(lv.STATE.CHECKED)

    fixture = self._invoke(self.cb1, active_index=-1)

    self.assertTrue(bool(self.cb1.get_state() & lv.STATE.CHECKED))
    self.assertEqual(fixture.active_radio_index, 1)
```

It exercises the exact scenario the checklist flags: a radio group with `active_radio_index == -1` (no prior selection) receives a click on an option. The code path that runs is the `else` branch of `radio_event_handler`, which this PR did **not** modify — only the `if not checked:` branch was touched.

Test run just now on macOS desktop:

```
./tests/unittest.sh tests/test_graphical_setting_activity_radio.py
...
test_click_active_option_keeps_it_selected ... ok
test_click_different_option_switches_selection ... ok
test_first_selection_from_empty_state ... ok
Ran 3 tests — OK
```

Also green across all 5 CI platforms on the latest run (Linux x64 / arm64 / special boards; MacOS Intel + arm64).

Ticking the box.

ThomasFarstrike:
Ah that's cool! One question I have though, is for things like the "Autostart at Boot" setting, does that mean that once a user has selected an app to start at boot, they will no longer be able to select "nothing" to start at boot?

Maybe this can be solved by adding an explicit "None" option at the top? Or some kind of "allow_deselect" parameter of a radio-button group, but I would be wary of "polluting" those parameters willy-nilly, as we might not want to add too many tiny various parameters that set tiny details of those settings, maybe...

bitcoin3us:
Great catch — you're right, the "Auto Start App" picker legitimately uses an empty-string value to mean "boot straight to the launcher", and the invariant in the previous version of this PR blocked users from ever reaching that state after picking something. Would've been a regression.

Between the two options you raised (add an explicit "None" radio button vs. a new `allow_deselect` flag) I went with the flag, for two reasons:

- For some settings a synthetic "None" option looks awkward and asks the app author to invent a label for "nothing". `Theme: Light / Dark / None?` is not great.
- The flag is opt-in with `default=False`, so every existing radio group (including all the others in the Settings tree, theme, timezone, audio device, wallet type, hero image, screen lock) keeps the new invariant without needing changes. Only settings that genuinely want the empty state raise their hand.

I agree about parameter creep and didn't want to add it lightly — but this is one well-motivated knob with a sensible default, not a sprawl of formatting options. Happy to reconsider if you'd still rather use the "None option" approach in the settings.py callers; I can rebase in that direction instead.

**What landed in commit `fc4df1fd`:**

1. **`lib/mpos/ui/setting_activity.py`** — when building a radio group, read `setting.get("allow_deselect", False)` and stash on the Activity as `self._radio_allow_deselect`. In `radio_event_handler`, on un-check of the active option, respect that flag: if True, allow the un-check and flip `active_radio_index` to -1 as before; if False, re-add `STATE.CHECKED` as this PR already did.

2. **`builtin/apps/com.micropythonos.settings/assets/settings.py`** — set `"allow_deselect": True` on the `Auto Start App` entry, with a comment explaining why.

3. **`tests/test_graphical_setting_activity_radio.py`** — new `test_allow_deselect_opt_in_lets_user_un_select_active` covering the opt-in path. Now 4 tests total, all pass locally:

```
test_click_active_option_keeps_it_selected ... ok
test_click_different_option_switches_selection ... ok
test_first_selection_from_empty_state ... ok
test_allow_deselect_opt_in_lets_user_un_select_active ... ok
Ran 4 tests — OK
```

4. **`CHANGELOG.md`** — extended the existing bullet to mention the opt-in and point to Auto Start App as the reference case.

I reviewed every other radio group in the tree: theme, timezone, audio default in/out, number format, restart-to-bootloader, wallet_type (Lightning Piggy), hero image, balance denomination, screen lock — all have "exactly one must be selected" semantics. So Auto Start App is currently the lone opt-in caller, which is the signal I was looking for that this wasn't an over-reach.

Let me know if you'd rather go the "None option" route and I'll rework.

PR #130, opened 2026-04-30
make expander analog method a property again

bertouttier:
The annotation was removed in 4b69252265d3b2ff97103d9f1e8674d14008001e


ThomasFarstrike:
Hmmm I think it was erroneously removed... should I add it again? Up to you, all the same for me.

bertouttier:
> Hmmm I think it was erroneously removed... should I add it again? Up to you, all the same for me.

I think it was indeed a mistake. This PR fixes it :) so feel free to merge.

PR #133, opened 2026-05-06
Add a driver and demo app for the DJ add-on

bertouttier:
Not fully finished yet, but good enough to add already.


bertouttier:
@ThomasFarstrike mock added, CI now passes. I think this should be good to go now!

ThomasFarstrike:
Awesome, thank you!

PR #134, opened 2026-05-08
add usb-device-midi from microptyhon-lib to the image

bertouttier:
The goal is to create a MicropythonOS app that acts as a USB MIDI device.


bertouttier:
Running the MIDI example in `lvgl_micropython/lib/micropython/lib/micropython-lib/micropython/usb/examples/device/midi_example.py` seems to work.

ThomasFarstrike:
This is very cool! It looks very small and will be frozen (compiled to .mpy) so it's fine, but in general, we might consider bundling these things with the app itself, instead of into MicroPythonOS, to keep the build small.

PR #136, opened 2026-05-13
DownloadManager: add redact_url= kwarg to hide auth-bearing URLs in logs

bitcoin3us:
## Summary

`DownloadManager.download_url` prints the full request URL three times per download (start, finished, exception path) plus the entire response-headers dict. Great for the typical callers fetching public URLs (app icons, OS updates, weather), but it silently leaks any auth secret embedded in the URL's path or query string.

Two real cases motivating this PR:

1. **Lightning Piggy's on-chain wallet** ([LightningPiggyApp PR #25](https://github.com/LightningPiggy/LightningPiggyApp/pull/25)) queries
   `https://btc1.trezor.io/api/v2/xpub/zpub6q…?details=txs&tokens=derived` to fetch balance + transactions for a watch-only xpub. The xpub itself sits in the URL path. A leaked xpub exposes the wallet's entire past + future address derivation tree to whoever reads the log — non-recoverable confidentiality damage even though funds custody is unaffected. The app already scrubs xpubs from its own `print()` calls, but `DownloadManager`'s prints sit below that and re-leak on every poll.

2. **Any future caller using API-key-in-URL or token-in-URL auth.** Common pattern (LNBits, BlueWallet's LNDHub, some HTTPS RPC endpoints) — the leak surface is wider than just Lightning Piggy.

## Behaviour

Adds a `redact_url=False` kwarg (default preserves current behaviour). When `True`:

- URL logged as `scheme://host[:port]/...REDACTED...` instead of full. **The host is intentionally kept** so failure triage (DNS / connectivity / wrong endpoint) is still possible.
- Response-headers dump is suppressed entirely (replaced with `<redacted>`). Response headers often carry `set-cookie`, `cf-ray`, and other correlatable session tokens.
- Exception messages have the URL substring scrubbed before printing (aiohttp's `ClientConnectorError` typically embeds the URL).

Default `False` is deliberate — most callers fetch public URLs and want the full log line for diagnostics. Callers explicitly opt in:

```python
await DownloadManager.download_url(url, redact_url=True)
```

## Tests

`tests/test_download_manager.py` — added 3 new test classes (10 tests):

- `TestSafeUrl` (6 tests) for the new `_safe_url` helper — path-with-query strip, port handling, naked-host pass-through, malformed-URL placeholder, belt-and-braces "secret substrings never appear in output" guard.
- `TestRedactUrlKwarg` (2 tests) confirming the kwarg flows through the call surface (sync wrapper + async impl + mock) and defaults to `False`.
- `TestRedactedExceptionPath` (2 tests, commit `64963f9`) — end-to-end coverage of the URL-bearing exception scrubbing path. Installs a fake `aiohttp` into `sys.modules` so the function-local `import aiohttp` resolves to a stub whose `ClientSession.get()` raises `RuntimeError(message)` with the full URL in the message (mimics aiohttp's `ClientConnectorError` without needing the network), then captures `print(...)` and asserts the secret-bearing substring is scrubbed under `redact_url=True` and preserved under `redact_url=False`.

All 22 existing tests continue to pass (default behaviour unchanged):

```
$ bash tests/unittest.sh tests/test_download_manager.py
Ran 32 tests
OK
```

## Follow-up

Lightning Piggy's on-chain wallet ([PR #25](https://github.com/LightningPiggy/LightningPiggyApp/pull/25)) will follow up on its side to pass `redact_url=True` once this lands. That's a one-line change in `onchain_wallet.py:fetch_balance_and_payments()`.

## Test plan

- [x] All 32 tests pass on macOS desktop (`bash tests/unittest.sh tests/test_download_manager.py`)
- [x] Existing tests (default behaviour) unchanged
- [x] **URL-bearing exception path covered by unit tests** (commit `64963f9`): fake-aiohttp injection deterministically exercises the `except Exception as e: ... url in err_str` branch and verifies both the scrubbed (`redact_url=True`) and preserved (`redact_url=False`) print output
- [ ] **Hardware end-to-end smoke test:** an LP onchain wallet call with `redact_url=True` shows `https://btc1.trezor.io/...REDACTED...` in serial logs. Gated on [LightningPiggy #25](https://github.com/LightningPiggy/LightningPiggyApp/pull/25) follow-up adding the kwarg — will tick once that lands and the integrated build is exercised on device

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Board Support:
 OS:
 - Disable the repl on hardware uart for esp32s3 targets (USB serial still works)
 
+Frameworks:
+- `DownloadManager.download_url`: add `redact_url=True` kwarg for callers fetching URLs that embed an auth secret (API key, OAuth token, LNBits readkey, xpub/ypub/zpub). When set, the URL is logged as `scheme://host/...REDACTED...`, the response-headers dump is suppressed, and exception messages have any embedded URL scrubbed. Default `False` preserves existing debug output for callers fetching public URLs (app icons, OS updates, etc.). Use case: prevents serial / REPL logs from leaking the secret-bearing URL even when DEBUG-level chatter is on.
+
 0.9.5
 =====
 
--- a/internal_filesystem/lib/mpos/net/download_manager.py
+++ b/internal_filesystem/lib/mpos/net/download_manager.py
@@ -28,12 +28,12 @@ class DownloadManager:
     @classmethod
     def download_url(cls, url, outfile=None, total_size=None,
                     progress_callback=None, chunk_callback=None, headers=None,
-                    speed_callback=None):
+                    speed_callback=None, redact_url=False):
         """Download a URL with flexible output modes (sync or async wrapper).
-        
+
         This method automatically detects whether it's being called from an async context
         and either returns a coroutine (for await) or runs synchronously.
-        
+
         Args:
             url (str): URL to download (required)
             outfile (str, optional): Path to write file. If None, returns bytes.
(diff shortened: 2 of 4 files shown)


bitcoin3us:
## CONTRIBUTING.md merge-checklist audit

Walking through each item explicitly so the maintainer doesn't have to re-derive it:

| Item | Status | Detail |
|---|---|---|
| 1. Expand `CHANGELOG.md` "Future release" | ✅ | Bullet added under Frameworks |
| 2. Increment app version in `META-INF/MANIFEST.JSON` | N/A | Framework change, not an app |
| 3. Update documentation repository | ✅ | [MicroPythonOS/docs#6](https://github.com/MicroPythonOS/docs/pull/6) — adds the new kwarg to the parameters table + a "Redacting Sensitive URLs" section covering when to opt in and what gets redacted |
| 4. Revise `MAINTAINERS.md` | N/A | No new boards |
| 5. Settings migration | N/A | No settings touched |
| 6. Unit tests | ✅ | 8 new tests in `tests/test_download_manager.py` covering the `_safe_url` helper and the kwarg's flow through both sync wrapper and async impl. All 30 download_manager tests pass |
| 7. Separate non-functional changes | ✅ | Single concern, no drive-by formatting |

AGENTS.md style:

| Rule | Status |
|---|---|
| ruff double-quotes | ✅ |
| Test placement (`mpos.testing` mocks where applicable) | ✅ Uses existing `MockDownloadManager` from `mpos.testing.mocks` |
| Test naming convention (`test_*.py`) | ✅ |

Companion docs PR: [MicroPythonOS/docs#6](https://github.com/MicroPythonOS/docs/pull/6) — ready to merge in lockstep with this one.

PR #148, opened 2026-05-30
Fast composed-font emoji rendering via lv_imgfont_set_range

bitcoin3us:
## Summary

Cuts the per-glyph cost of `FontManager.getFont(emoji=True)` labels from a C→Python round-trip per glyph to two int compares in C, for any codepoint that isn't in the emoji map. User-visible result: noticeably smoother scrolling of any label that uses an emoji-fallback font (Lightning Piggy transactions row, Files list, etc.) — even though most rows have no emoji in them.

## How

Three pieces:

* **`internal_filesystem/lib/mpos/ui/font_manager.py`** — `_create_emoji_font(size)` now calls `lv.imgfont_set_range(font, cp_min, cp_max, 0xE000, 0xF8FF)` after `lv.imgfont_create`. `cp_min`/`cp_max` are computed dynamically from the loaded `_emoji_maps` keys, so the filter self-maintains as the emoji set grows — no hardcoded magic number. Guarded by `try/except AttributeError` so older `lvgl_micropython` pins still work (just without the speedup).
* **`scripts/build_mpos.sh`** — applies `lvgl_micropython/imgfont_set_range.patch` against `lib/lvgl` during the build. Gated by a `-f` existence check, so older `lvgl_micropython` pins (no patch file present) still build cleanly.
* **`lvgl_micropython` submodule** — bumped `916ec6b` → `e00d9d9`. Picks up the new `imgfont_set_range.patch` plus a small drive-by brew-config parser fix in `builder/macOS.py` that was breaking macOS builds entirely.

## Dependency

This PR depends on MicroPythonOS/lvgl_micropython#3. The submodule pointer here references `e00d9d9`, which currently lives only on the `bitcoin3us/lvgl_micropython` fork. Once #3 merges into `MicroPythonOS/lvgl_micropython`, this PR's submodule pointer will resolve from upstream and this can land. `.gitmodules` is intentionally untouched (still points at the canonical URL).

## Test plan

- [x] macOS desktop build (`build_mpos.sh macOS`) — clean, `lv.imgfont_set_range` exposed in the LVGL binding
- [x] ESP32-S3 firmware build (`build_mpos.sh esp32s3`) — clean build, patch applies, all expected freezing happens
- [x] Flash + boot on Waveshare ESP32-S3-Touch-LCD-2 — boots cleanly, `lv.imgfont_set_range` callable, FontManager wiring works
- [x] FontManager bounds calculation on device — with the 400-emoji set loaded, bounds resolve to `(0x203C, 0x1F9E2)` (exact min/max across the loaded codepoints, as designed)
- [x] Backward-compat fallback — `try/except AttributeError` path verified mentally (covered by gating in build_mpos.sh)
- [x] User test — scrolling the Lightning Piggy transactions list feels noticeably smoother than the pre-patch firmware

### DIFF
--- a/internal_filesystem/lib/mpos/ui/font_manager.py
+++ b/internal_filesystem/lib/mpos/ui/font_manager.py
@@ -280,9 +280,55 @@ birthday_cakes,👑
         cls._ensure_emoji_maps()
 
         try:
-            return lv.imgfont_create(size, cls._imgfont_path_cb, None)
+            font = lv.imgfont_create(size, cls._imgfont_path_cb, None)
         except Exception:
             return None
+        if font is None:
+            return None
+
+        # Push the same codepoint accept/exclude ranges that _get_emoji_src
+        # checks down into the C-level imgfont. Once set, LVGL stops calling
+        # our Python _imgfont_path_cb for codepoints that are guaranteed to
+        # have no emoji glyph (ASCII, CJK, PUA, etc.) — turning the per-glyph
+        # cost from "Python call + return" into a pair of int compares in C.
+        # This is the actual fix for the scrolling slowdown: composed-font
+        # text where most glyphs are non-emoji now stays in C end-to-end.
+        try:
+            cp_min, cp_max = cls._emoji_codepoint_bounds()
+            lv.imgfont_set_range(font, cp_min, cp_max, 0xE000, 0xF8FF)
+        except AttributeError:
+            # Older LVGL build without lv_imgfont_set_range — that's OK, the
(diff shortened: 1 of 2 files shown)


ThomasFarstrike:
That's very impressive! I like how the range can be set at runtime, so it's more generic. Not sure upstream lvgl_micropython or upstream lvgl_micropython/lib/lvgl would ever merge this, but that's not a dealbreaker!

Let me merge it, assuming you've already verified that it works. I'm excited!

PR #164, opened 2026-06-16
FontManager: add 11 emojis to the built-in pack (incl. 🇸🇻, 🏎️, 🌽, 🍕)

bitcoin3us:
Adds 13 Noto 32×32 emoji PNGs to `internal_filesystem/builtin/res/emojis/32x32/`, which `FontManager._build_emoji_map` auto-discovers (same mechanism as the recent "Smiling Face with Three Hearts" addition).

## Added

🏎️ racing car · 🌽 corn · 🍕 pizza · 💨 dashing-away · 💥 collision · 🚽 toilet · 😷 face with medical mask · ✊ raised fist · 🫶 heart hands · 🧡 orange heart · 💜 purple heart · 🐦 bird · 🇸🇻 El Salvador flag

## Sequence handling

- **🇸🇻 flag** resolves via the existing 2-codepoint combine in `_imgfont_path_cb` (the 2nd regional indicator `1F1FB` is a modifier), keyed `1F1F8-1F1FB.png`. Sourced from Noto's region-flag and letterboxed into a 32×32 transparent square — noto-emoji doesn't ship the stylized waved-flag PNG individually, so it renders flat rather than the rounded-emoji look. Trivial to swap if you have a stylized source.
- **🤷‍♂️** (`1F937 200D 2642 FE0F`) gracefully falls back to the existing neutral shrug `1F937.png` — no new file needed.
- **🐦‍⬛** (black bird) isn't a clean 2-part combine, so it renders as the plain bird `1F426` added here; the `⬛` doesn't merge.

Already present, so skipped: 💩 `1F4A9`, 🚀 `1F680`, ✨ `2728`.

## Verification

- The emoji map resolves every new key, including the flag sequence and the pre-existing fallbacks — checked by replicating `_build_emoji_map` over the dir.
- Source glyphs spot-checked (correct corn / flag / car / etc.).
- `showfonts` demo updated to display all the new glyphs for on-target visual checking.

## Merge-checklist notes

- **CHANGELOG**: entry added under Future release → Frameworks.
- **Tests**: none — this is an asset addition; rendering is exercised by the showfonts demo. Not run on a built target here (no local emoji-build), so a quick look in showfonts after building is worth it.
- **PNG optimization**: done — each PNG was run through `optipng -o7` + `zopflipng --lossy_8bit --lossy_transparent` (the same flags as `scripts/optimize_png.sh`), shrinking the 13 files from 26.4 KiB to 18.0 KiB total (−31%), in line with the rest of the pack. Glyphs re-checked after the lossy pass.
- **Licensing**: Noto emoji (Apache-2.0), consistent with the existing pack / `NotoColorEmoji` license already in the tree.


ThomasFarstrike:
Haha, cool!

Note that we have to try and conserve space as much as possible, so try not to add too many emojis... 

May I ask the reasons to add these? Like, why the face mask? Is that used often?
Same question for the bird...
I guess the toilet, collision and fart-like "dashing-away" are for kids with their toilet humor? :-)

I don't mean to be strict, to me those are all fine, it just helps to have a little rationale when we add (or remove) emojis, fur future reference.

bitcoin3us:
Totally fair on size — let's keep it tight. **Dropped 😷 face mask and 🚽 toilet** (pushed), which trims the net add from 13 to **11**.

Rationale for the rest, for the record:

- **🏎️ racing car** — used by D++ in her *Satoshi Speedway* Mario Kart Lightning-stream messages
- **🌽 corn** — going to be added as a unit of bitcoin in the Lightning Piggy app
- **🍕 pizza** — Bitcoin Pizza Day
- **💨 dashing-away** — a nod to farts *and* speed
- **🧡 orange heart** — Bitcoin love
- **💜 purple heart** — Nostr love
- **🐦 bird** — Sparrow wallet
- **🇸🇻 El Salvador flag** — the saviour

And these three I use personally and was a bit surprised aren't more common: **💥 collision · ✊ raised fist · 🫶 heart hands**

(As noted earlier in the thread, 💩/🚀/✨ were already in the pack, and 🤷‍♂️ falls back to the existing neutral shrug, so those aren't new files.)

ThomasFarstrike:
Did it work properly for you? Because the El Salvador flag was not showing up for me...

**Root cause**

The El Salvador flag is the first regional-indicator flag emoji in the set (🇸 🇻 = U+1F1F8 U+1F1FB). In internal_filesystem/lib/mpos/ui/font_manager.py, regional indicators are classified as “emoji modifiers” so they don't render as standalone glyphs. However, the modifier early-return fired before the two-letter pair could be looked up, so both characters rendered as empty/transparent images.

**Fix**

- Added _is_regional_indicator() to detect regional indicator codepoints.
- Added explicit flag-pair handling in _imgfont_path_cb() before the general modifier early-return: when one regional indicator is followed by another, it looks up the combined key like 1F1F8-1F1FB in the emoji map and renders the flag. Unpaired regional indicators still return the empty placeholder.
- Confirmed the lvgl_micropython/ patches (e.g. imgfont_set_range.patch) are not the cause — the issue was entirely on the MicroPythonOS FontManager side.

**Tests**

Added TestFontManagerFlagEmoji in tests/test_graphical_font_manager.py with regression tests:
- regional indicator detection
- flag pair resolves to a non-empty scaled image descriptor
- unpaired regional indicator returns the 1px empty descriptor
- El Salvador flag renders in a label
- all 11 newly added emojis render in a label


Anyway, no action required; I added the 11 emojis you proposed based on this pull request :-)

ThomasFarstrike:
Merged! I did it manually because there were some conflicts :-)

bitcoin3us:
> Merged! I did it manually because there were some conflicts :-)

Thank you. Apologies for not responding sooner. Just picking up your reply now.

PR #166, opened 2026-06-18
fri3d 2026: add new coprocessor firmware 2.0.0

bertouttier:
This new firmware is needed for the production version of the badge. It has the following changes:
* removed 1 analog DMM channel
* added support to control the Lora module reset line (done at boot together with the LCD)
* charger standby pin reading is fixed


bertouttier:
Idea: create a subclass of `class SX126X` and override the `def __init__()` and `def reset()` functions to use the `expander.config` to trigger a reset. Downside: controlling the value of `expander.config` in multiple locations might cause troubles.

ThomasFarstrike:
Another note is that we currently carry 2 versions of that SX1262 driver:
- internal_filesystem/lib/drivers/lora/sx1262.py # used by fri3d_2026 
- internal_filesystem/lib/drivers/lora/micropySX126X # used by T-Watch-S3 Plus

And they might be out of sync with https://github.com/micropython/micropython-lib/tree/master/micropython/lora so in general, this could use some consolidation...

PR #167, opened 2026-06-19
revert register map of the expander and add firmware version v2.0.1

bertouttier:
* adds coprocessor firmware v2.0.1 which adds a backwards compatibilty fix.
* reverts the expander driver to the original register map


ThomasFarstrike:
Nice one!

PR #170, opened 2026-06-24
fri3d_2026: fix SD card and Lora clash

bertouttier:
* set the SPI bus frequency to 16MHz for Lora, Display and SD card
* Add the Lora Chip Select pin

### DIFF
--- a/internal_filesystem/lib/mpos/board/fri3d_2026.py
+++ b/internal_filesystem/lib/mpos/board/fri3d_2026.py
@@ -49,6 +49,7 @@ LightsManager.set_led_num(5)
 LightsManager.set_led(4, 21, 96, 67)
 LightsManager.write()
 
+spi_freq = const(16000000)
 spi_bus = SPI.Bus(
     host=2,
     mosi=6,
@@ -58,7 +59,7 @@ spi_bus = SPI.Bus(
 
 # Would be better to do this only when the LoRa app starts:
 try:
-    lora_spi_device = SPI.Device(spi_bus=spi_bus, freq=500000, cs=-1, polarity=0, phase=0, firstbit=SPI.Device.MSB, bits=8)
+    lora_spi_device = SPI.Device(spi_bus=spi_bus, freq=spi_freq, cs=45, polarity=0, phase=0, firstbit=SPI.Device.MSB, bits=8)
 except Exception as e:
     import sys
     sys.print_exception(e)
@@ -73,7 +74,7 @@ else:
 
 display_bus = lcd_bus.SPIBus(
     spi_bus=spi_bus,
-    freq=40000000, # 40 Mhz
+    freq=spi_freq, # 40 Mhz
     dc=4,
     cs=5
 )
@@ -230,7 +231,7 @@ except Exception as e:
     logger.error("expander init got exception: %s" % (e))
 
 import mpos.sdcard
-mpos.sdcard.init(spi_bus=spi_bus, cs_pin=14)
+mpos.sdcard.init(slot=2, spi_bus=spi_bus, cs_pin=14, freq=spi_freq)
 
 IRManager.txPin = Pin(21, Pin.OUT) # mini blaster / noisycricket has an IR LED
 IRManager.rxPin = Pin(11, Pin.IN)


ThomasFarstrike:
Awesome that you found a fix!!!

I'm just wondering whether 16Mhz isn't so low that it will make the display slow. I seem to remember playing around with it and finding that the framerate suffered or there were huge "tearing" effects or something. But I haven't tried since I increased the display's internal refresh rate to 90Hz instead of the previous 30Hz.

May I ask how you ended up going for 16Mhz? Is that the highest possible? Or an educated guess that turns out to work?

Thanks!

bertouttier:
> May I ask how you ended up going for 16Mhz? Is that the highest possible? Or an educated guess that turns out to work?

I looked at the maximum value for the SX1262 mainly (It should be 16MHz according to the datasheet). But I can do more testing before we merge this!

ThomasFarstrike:
Hmmm, I think I notice a lower framerate, just subjectively seeming to notice more stutter while scrolling.

Also, it (usually) doesn't seem to work.

Without an SD card plugged:

```
>>> LoRa Chat foregrounded, starting receive_thread
starting lora in 1 second
RF_SW set to HIGH
Unhandled exception in thread started by <bound_method>
  File "apps/com.micropythonos.lora_chat/lora_chat.py", line 174, in receive_thread
  File "lib/drivers/lora/sx1262.py", line 2165, in begin
  File "lib/drivers/lora/sx1262.py", line 407, in ASSERT
AssertionError: ERR_SPI_CMD_TIMEOUT
```

With a 64 GB SD card is plugged in, I instead get:

```
LoRa Chat foregrounded, starting receive_thread
starting lora in 1 second
RF_SW set to HIGH
Unhandled exception in thread started by <bound_method>
  File "apps/com.micropythonos.lora_chat/lora_chat.py", line 173, in receive_thread
  File "lib/drivers/lora/sx1262.py", line 2153, in begin
  File "lib/drivers/lora/sx1262.py", line 515, in begin
  File "lib/drivers/lora/sx1262.py", line 407, in ASSERT
AssertionError: ERR_WRONG_MODEM
```

or:

```
LoRa Chat foregrounded, starting receive_thread
starting lora in 1 second
RF_SW set to HIGH
Unhandled exception in thread started by <bound_method>
  File "apps/com.micropythonos.lora_chat/lora_chat.py", line 173, in receive_thread
  File "lib/drivers/lora/sx1262.py", line 2153, in begin
  File "lib/drivers/lora/sx1262.py", line 509, in begin
  File "lib/drivers/lora/sx1262.py", line 407, in ASSERT
AssertionError: ERR_CHIP_NOT_FOUND
```

ThomasFarstrike:
Did it work fine for you?

In that case, maybe the explicit LoRa reset handling is fixing it. That would be good news!

I read that "30 MHz across a 320 × 240 screen limits you to roughly 20 FPS" so 16Mhz is probably around 10 FPS, which is also what it looks like... coming from ~25 FPS, that's a real downgrade...

Hopefully you can get all 3 (display, SD card and LoRa) working at 40 Mhz.

Or, we only set the SPI to that 16 Mhz clock speed if there's a LoRa chip detected?

To me it's strange that the current version (which works fine except when an SD card is plugged in) has 500 kHz for LoRa and 40 Mhz for the display. How does that work? I can't imagine it often reconfigures the clock at runtime?

ThomasFarstrike:
I set this to "Draft" because I understand it's not very stable yet...?

ThomasFarstrike:
@bertouttier did you see https://github.com/MicroPythonOS/MicroPythonOS/pull/222 ?

bertouttier:
> @bertouttier did you see https://github.com/MicroPythonOS/MicroPythonOS/pull/222 ?

Yes, I did, but not the fix in detail :) Claude gave me the same feedback though, so I was looking in the same direction.

ThomasFarstrike:
1) I've applied your LoRa frequency change from 500 kHz to 16 Mhz and that seems to work well.
Along with https://github.com/MicroPythonOS/MicroPythonOS/pull/222 for stability, it's a nice improvement.
Thanks!

2) The SD card frequency is fine being handled by ESP-IDF (first 400 kHz for init, then auto negotiated up to 20 Mhz) so no need to tweak it.

3) Display frequency needs to be high to have decent framerate, so let's leave it at 40 Mhz unless there's a compelling reason to lower it.

With that, I think we can close this pull request? Feel free to comment if it needs reopening!

PR #173, opened 2026-06-25
Focus: defer focus border until first directional navigation

bitcoin3us:
Keep focus borders hidden until the user actually navigates by direction (joystick/keypad).

## Why
`add_focus_border()` draws a focus highlight and adds the widget to the default focus group. That highlight is a navigation aid for **keypad/encoder** devices, showing which widget the moving focus is on. On a **pointer/touch** UI there is no moving highlight: the ring only appears after a tap and then lingers around the tapped widget.

Concretely, this is why the Lightning Piggy balance shows a gold rectangle after you tap it (the balance is `add_focus_border`-ed, and tapping it cycles the denomination). On a touchscreen that ring is visual noise rather than a cue.

## What
Per review feedback below, this is **input-driven rather than capability-driven**. Every widget is still added to the focus group; only the drawing of the border is deferred:

- `mpos/ui/focus.py`: module flag `_focus_nav_active` (starts `False`) plus `enable_focus_borders()`. `_focus_border_handler()` returns early while the flag is `False`, so no border is drawn.
- `mpos/ui/focus_direction.py`: `move_focus_direction()` calls `enable_focus_borders()` at the top, so the first joystick/keypad press flips it on and the resulting `FOCUSED` event draws the border as normal.

Resulting behavior:

| device | result |
| --- | --- |
| touch only (focus by tapping) | never shows a ring |
| keypad/encoder | ring appears as soon as they navigate |
| hybrid touch + keypad | ring appears only once the joystick is actually used |

The hybrid case is the reason for this approach over the first version of this PR, which skipped `add_focus_border()` entirely when a pointer indev was present and so would have wrongly suppressed the ring on hybrid devices. `add_focus_border()` itself is back to its original shape, with no capability check, making this a smaller change than that first version.

## Testing
- `tests/test_focus_border_deferred.py` added: asserts no border is drawn before directional navigation, and that a border is drawn after `enable_focus_borders()`.
- Verified both states in a CPython stub harness (lvgl stubbed): passes.
- Caveat, stated plainly: I did **not** run `./tests/unittest.sh` locally, since it needs a freshly built desktop binary and my local build is stale. Please run `./tests/unittest.sh tests/test_focus_border_deferred.py` in CI or on a current build.

## Merge checklist
- CHANGELOG: added under Future release, Frameworks
- App version / MANIFEST: n/a (framework change)
- MAINTAINERS.md / boards: n/a
- Settings migration: n/a
- Unit tests: added, see the testing caveat above
- Frameworks docs (docs.micropythonos.com): not updated. `add_focus_border` behavior is now state-dependent, so happy to add a note if there is a docs page covering it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Builtin Apps:
 - OSUpdate: show raw ESP error and a user-friendly message when the update fails to activate
 
 Frameworks:
+- Focus: `add_focus_border()` highlights now stay hidden until the user navigates by direction (joystick/keypad). The widget is still added to the focus group, but the border is only drawn from the first `move_focus_direction()` onward. Touch-only UIs (which focus widgets by tapping) never show a stray ring — e.g. the gold rectangle that appeared around the Lightning Piggy balance after tapping it — while keypad/encoder navigation is unchanged.
 - AppManager: add support for apps as Python packages 
 - Strip trailing slashes from directory paths before filesystem calls to support both LittleFS and FAT32 (SD card)
 - SDCard: tolerate trailing slashes when listing SD card contents
--- a/internal_filesystem/lib/mpos/ui/focus.py
+++ b/internal_filesystem/lib/mpos/ui/focus.py
@@ -1,7 +1,26 @@
 import lvgl as lv
 
 
+# Focus borders highlight which widget directional (keypad/joystick) focus is
+# on. They must stay invisible until the user actually navigates by direction,
+# so a touch-only UI — where widgets are focused by tapping — never shows a
+# stray highlight, and a hybrid touch+keypad device only shows it once the
+# joystick is used. move_focus_direction() flips this on first use via
+# enable_focus_borders().
+_focus_nav_active = False
+
+
+def enable_focus_borders():
+    """Mark that the user has navigated by direction (joystick/keypad). From
+    then on, focused widgets draw their focus border."""
+    global _focus_nav_active
+    _focus_nav_active = True
+
+
 def _focus_border_handler(event, width, color, opacity, radius):
+    if not _focus_nav_active:
+        # No directional navigation yet (e.g. touch-only use): no highlight.
+        return
     target = event.get_target_obj()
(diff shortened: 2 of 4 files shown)


ThomasFarstrike:
Would indeed be nice to find a solution for this and I guess it would work but indeed:

> Hybrid devices (both touch and a keypad) would also skip the ring

...but I was thinking, how about on all devices, we still add the widget to the focusgroup, but don't do anything with the border (no set_style_border_color etc) *until* the user uses "move_focus_direction" , meaning they pushed the joystick?

So first they don't see anything special, but as soon as they press the joystick *once* the set_style_border_color , set_style_border_width , set_style_border_opa and set_style_radius are applied so it behaves as normal?

bitcoin3us:
Good call — adopted your approach. Pushed a rewrite (force-update, rebased onto current `main`, so the CHANGELOG conflict is resolved).

Instead of skipping `add_focus_border()` on touch devices, every widget is still added to the focus group, but the border is now **deferred until the first directional navigation**:

- `mpos/ui/focus.py`: module flag `_focus_nav_active` (starts `False`) + `enable_focus_borders()`. `_focus_border_handler()` early-returns while the flag is `False`, so no border is drawn.
- `mpos/ui/focus_direction.py`: `move_focus_direction()` calls `enable_focus_borders()` at the top, so the first joystick/keypad press flips it on and the resulting `FOCUSED` event draws the border as normal.

Behavior:
- **touch-only** (focus by tapping) → never shows a ring;
- **keypad/encoder** → shows it the moment they navigate;
- **hybrid** → shows it only once the joystick is actually used — which fixes the caveat you flagged.

`add_focus_border` itself is back to its original shape (no capability check), so it's a smaller change than my first version.

Testing: `tests/test_focus_border_deferred.py` asserts no border before nav and a border after `enable_focus_borders()`. Verified the logic in a CPython stub (both states pass); I did **not** run `./tests/unittest.sh` locally (needs a fresh desktop build), so please run it in CI / on a current build.

PR #177, opened 2026-06-28
feat(net): off-loop DNS resolution so connecting doesn't freeze the event loop / UI

jnuyens:
## Summary

Opening a network connection briefly **freezes the whole UI** during the DNS
lookup (worst for unreachable / non-existent hosts). Easy to reproduce by
connecting to a random non-existent host (e.g. `akhegfkfkqhskfhesbfkwlmd.com`).

### Root cause

`asyncio.open_connection()` calls `socket.getaddrinfo()` **synchronously before
its first `yield`** (`extmod/asyncio/stream.py` -- the upstream source even has a
`# TODO this is blocking!` there). On ESP32-S3 that blocking DNS call runs on the
asyncio event-loop thread, which is also the thread that drives the LVGL UI via
`micropython.schedule()`. So while `getaddrinfo` waits, the scheduler can't run,
LVGL never ticks, and the UI is frozen for the full DNS duration.

This is **not** an unfixable ESP-IDF limitation: lwIP DNS is async internally, and
ESP32's `getaddrinfo` releases the GIL while it waits (`ports/esp32/modsocket.c`),
so another `_thread` keeps running during the lookup. The fix is entirely at the
Python layer.

## Fix

- New `mpos/net/async_dns.getaddrinfo_async()`: resolves DNS in a `_thread` worker
  (same pattern as `audiomanager.py` / `wifi_service.py`) and `await`-polls via
  `TaskManager.sleep_ms()`, so the event loop / UI keep ticking while DNS resolves.
  It does not import `asyncio` (honors the "only TaskManager imports asyncio" rule).
- `aiohttp` now resolves the host off-loop, then connects to the resolved IP with
  `server_hostname=host` so **TLS SNI and cert validation still use the real
  hostname**. DNS + connect run together under the existing request `timeout`.

## Testing

- `tests/test_async_dns.py` (network-free, deterministic): proves the loop keeps
  ticking while a 150 ms-blocking lookup runs, and that worker exceptions propagate.
- Regression: existing aiohttp-path tests still pass.
- Live-verified on the desktop simulator against a dead host: old path = **0** UI
  ticks during the lookup; new path = loop kept ticking the whole time. (On-device
  the difference is multi-second.)

No new dependencies; ASCII-only; behavior unchanged for normal hosts.

### DIFF
--- a/internal_filesystem/lib/aiohttp/__init__.py
+++ b/internal_filesystem/lib/aiohttp/__init__.py
@@ -3,6 +3,7 @@
 
 import asyncio
 import json as _json
+from mpos.net.async_dns import getaddrinfo_async
 from .aiohttp_ws import (
     _WSRequestContextManager,
     ClientWebSocketResponse,
@@ -250,12 +251,24 @@ class ClientSession:
             host, port = host.split(":", 1)
             port = int(port)
 
+        async def _resolve_and_connect():
+            # Warm the DNS resolver off the event-loop thread so the UI stays
+            # responsive during the (possibly slow) lookup: getaddrinfo releases the
+            # GIL while it waits (modsocket.c:239-253), so the worker thread keeps the
+            # loop alive. open_connection() below then resolves from the resolver
+            # cache (fast) instead of blocking the loop.
+            #
+            # We pass the hostname (not a resolved address) to open_connection so TLS
+            # SNI and cert validation use the real host. MicroPython getaddrinfo
+            # returns the sockaddr as an opaque bytearray (not an (ip, port) tuple),
+            # so the address cannot be substituted portably anyway.
+            await getaddrinfo_async(host, port)
+            return await asyncio.open_connection(host, port, ssl=ssl)
+
         if timeout is not None:
-            reader, writer = await asyncio.wait_for(
-                asyncio.open_connection(host, port, ssl=ssl), timeout
-            )
+            reader, writer = await asyncio.wait_for(_resolve_and_connect(), timeout)
         else:
-            reader, writer = await asyncio.open_connection(host, port, ssl=ssl)
+            reader, writer = await _resolve_and_connect()
 
         # Use protocol 1.0, because 1.1 always allows to use chunked transfer-encoding
         # But explicitly set Connection: close, even though this should be default for 1.0,
--- a/internal_filesystem/lib/mpos/net/async_dns.py
+++ b/internal_filesystem/lib/mpos/net/async_dns.py
@@ -1,0 +1,114 @@
+"""async_dns.py -- Off-loop DNS resolution helper for MicroPythonOS.
+
+Offloads socket.getaddrinfo to a _thread worker so the asyncio event loop
+stays alive during blocking DNS lookups on ESP32-S3.
+
+On ESP32-S3, socket.getaddrinfo() is a blocking call that wraps lwIP's DNS
+resolver (lwip_getaddrinfo in ports/esp32/modsocket.c). Although lwIP DNS is
+internally async, the MicroPython wrapper blocks the calling thread until the
+result arrives. When this call runs on the asyncio event-loop thread (which is
+also the LVGL task_handler thread), the entire UI freezes for the duration of
+the DNS lookup -- worst case several seconds for non-existent hosts.
+
+The GIL is released during the blocking C call (modsocket.c:239-253), so a
+_thread worker keeps the main loop alive while DNS resolves. This module
+provides getaddrinfo_async(), an awaitable coroutine that spawns the blocking
+getaddrinfo in a worker thread and polls for completion with 20 ms sleeps,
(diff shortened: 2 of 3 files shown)


ThomasFarstrike:
This is very hopeful!

What about the failing unit tests, do the tests need updating? Or is something still wrong?

For example: https://github.com/MicroPythonOS/MicroPythonOS/actions/runs/28320295078/job/83900969795?pr=177

```
935710] Closing ClientSession
935710] _async_main completed
Error: 935712] _async_main's await self._connect_and_run() for wss://relay.0xchat.com got exception: can't convert 'int' object to str implicitly
935712] No reconnect configured, breaking loop
935712] Initiating cleanup
relay.py on_close
Error: 935712] _async_main's await self._connect_and_run() for wss://relay.primal.net got exception: can't convert 'int' object to str implicitly
935712] No reconnect configured, breaking loop
935712] Initiating cleanup
relay.py on_close
935712] Callback task cancelled
935712] Closing WebSocket connection
935712] WebSocket already closed or not initialized
935712] Closing ClientSession
935712] Callback task cancelled
935712] Closing WebSocket connection
935712] WebSocket already closed or not initialized
935712] Closing ClientSession
935712] _async_main completed
935712] _async_main completed
NostrManager: 3 relay(s) connected
publishing message to relays: {'wss://relay.damus.io': <Relay object at 7fa09ca50dc0>, 'wss://relay.0xchat.com': <Relay object at 7fa09ca51000>, 'wss://relay.primal.net': <Relay object at 7fa09ca50fa0>}
NostrManager: subscribed to 'nip17-debug'
publishing message to relays: {'wss://relay.damus.io': <Relay object at 7fa09ca50dc0>, 'wss://relay.0xchat.com': <Relay object at 7fa09ca51000>, 'wss://relay.primal.net': <Relay object at 7fa09ca50fa0>}
publishing message to relays: {'wss://relay.damus.io': <Relay object at 7fa09ca50dc0>, 'wss://relay.0xchat.com': <Relay object at 7fa09ca51000>, 'wss://relay.primal.net': <Relay object at 7fa09ca50fa0>}
NostrManager: published relay lists (2)
 FAIL

======================================================================
FAIL: test_self_send_nip17_over_real_relays <class 'TestNip17SelfSendRealRelays'>
----------------------------------------------------------------------
  File "lib/unittest/__init__.py", line 398, in run_one
  File "<stdin>", line 163, in test_self_send_nip17_over_real_relays
  File "<stdin>", line 68, in _run_async
  File "asyncio/core.py", line 1, in run
  File "asyncio/core.py", line 1, in run_until_complete
  File "asyncio/core.py", line 1, in run_until_complete
  File "<stdin>", line 127, in _async_test_self_send
  File "lib/unittest/__init__.py", line 176, in assertTrue
AssertionError: NostrManager did not connect to any relay within 30s

----------------------------------------------------------------------
Ran 1 tests

FAILED (failures=1, errors=0)
~/work/MicroPythonOS/MicroPythonOS



WARNING: test /home/runner/work/MicroPythonOS/MicroPythonOS/tests/test_nostr_nip17_self_send_real_relays.py got error 1 !!!
```

ThomasFarstrike:
While the same tests on the main branch went fine: https://github.com/MicroPythonOS/MicroPythonOS/actions/runs/28318029193

ThomasFarstrike:
Could also be you just need to re-sync with the main branch, of course... including the micropython-nostr/ branch.

jnuyens:
Good catch -- you were right, something *was* still wrong, and it was my bug (not a re-sync issue).

**Root cause:** my first version extracted the resolved IP via `ai[0][-1][0]` and connected to that. That works on CPython, but in MicroPython `socket.getaddrinfo()` returns the sockaddr as an opaque **`bytearray`**, not an `(ip, port)` tuple -- so `ai[0][-1][0]` was the first *byte* of the struct (an int), and `open_connection(<int>, ...)` raised `TypeError: can't convert 'int' object to str implicitly`, breaking every TLS connection. My unit test mocked `getaddrinfo` with CPython-style tuples, so it didn't catch it -- but `tests/test_websocket.py` did (on_open went 3 -> 0). Thanks for flagging it.

**Fix (just force-pushed):** don't substitute the address at all. Resolve DNS off-loop only to *warm the resolver*, then call the original `open_connection(host, port, ssl=ssl)` -- the follow-up lookup is a cache hit, the hostname is preserved for TLS SNI, and there's no fragile sockaddr parsing. Locally now: `test_websocket.py` on_open=3 / OK, `test_appstore_async_refresh.py` 11/11, plus the new `test_async_dns.py`.

The remaining macOS build failures look like the separate Homebrew `sdl2` 404 (`sdl2` -> `sdl2-compat`) you're already tackling on `main` -- unrelated to this PR.

ThomasFarstrike:
It's a bit dangerous to open new threads because we're very limited by on-board RAM but the max limit is good and we can adjust it downward if needed... let's see how it goes!

PR #178, opened 2026-06-28
fix(audio): stop desktop player by PID, not pkill -f (no cross-kill)

jnuyens:
Follow-up to the desktop-audio player (#175).

## Problem
The desktop (non-ESP32) `WAVStream` branch backgrounded an external player
(`afplay`/`ffplay`/`aplay`/`paplay`) and stopped it with:

```python
os.system("pkill -f %s" % _shell_quote(self.file_path))
```

`pkill -f` matches **any** process whose command line contains that path, so it
can kill unrelated processes, or cross-kill a second player of the same file.

## Fix
Record the backgrounded player's PID via the shell's `$!` into a per-stream pid
file at launch, and stop it by that exact PID:

```python
cmd = "... >/dev/null 2>&1 & echo $! > <pidfile>"
...
_stop_desktop_player(pid_file, not self._keep_running)   # kill <pid> on explicit stop
```

Kill happens only on an explicit `stop()`; on natural completion the player has
already exited (so we just remove the pid file — which also avoids PID reuse).

## Testing
Verified on the macOS sim binary: a bystander process sharing the **same file
path** survives the stop (it would have been killed by the old `pkill -f`), the
real player is killed by PID, and the pid file is cleaned up. ESP32 I2S path
untouched; `mpy-cross` compiles; ASCII-only.

### DIFF
--- a/internal_filesystem/lib/mpos/audio/stream_wav.py
+++ b/internal_filesystem/lib/mpos/audio/stream_wav.py
@@ -43,6 +43,27 @@ def _shell_quote(path):
     return "'" + path.replace("'", "'\"'\"'") + "'"
 
 
+def _stop_desktop_player(pid_file, kill):
+    """Stop a backgrounded desktop player by its recorded PID.
+
+    Killing the exact PID (vs `pkill -f <path>`) avoids matching unrelated
+    processes or cross-killing another player of the same file. Only called to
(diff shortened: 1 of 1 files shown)


ThomasFarstrike:
Looks good, thank you!

PR #179, opened 2026-06-28
fix(net): aiohttp header-read timeout + mutable-header poisoning + atomic wifi_busy

jnuyens:
A small batch of networking robustness fixes (maintainer shortlist).

### 1. aiohttp: stalled server can hang the loop/UI forever
`_request`'s timeout covered only DNS+connect; the status-line and response-header
`readline()` reads were unbounded. A server that accepts the connection then stalls
before/while sending headers froze the asyncio event loop (and the LVGL UI) forever.
The header read is now wrapped in the same `wait_for(timeout)`. Body reads remain the
caller's responsibility (`DownloadManager` already bounds per-chunk reads).

### 2. aiohttp: mutable-default `headers={}` poisoning
The four `headers={}` default parameters are now `headers=None`, and `request_raw`
copies the dict before mutating it (`Host`/`Content-Type`/`Content-Length`) — so a
caller-shared headers dict can no longer be poisoned across requests.

### 3. WifiService: `wifi_busy` check-then-set race
`auto_connect()` and `temporarily_disable()` each tested `wifi_busy` then set it in
separate statements. Two threads (boot auto-connect + the WiFi app) could both see
`wifi_busy=False` and proceed → double-connect. Added a `_busy_lock` + `_acquire_busy()`
helper that does the check-and-set atomically, used at both sites.

### Testing (macOS sim)
- New stall test: with a server that connects but never sends headers, `_request(timeout=1)`
  now raises `TimeoutError` at ~1.0s instead of hanging (verified old path would hang).
- Header isolation: two `ClientSession`s with different default headers stay isolated.
- Existing suites: `test_websocket` OK, `test_async_dns` OK, `test_wifi_service` 46/46 OK.
- `mpy-cross` compiles both files on this branch's base; ASCII-only.

Independent of #177 (touches different regions of `aiohttp/__init__.py`).

### DIFF
--- a/internal_filesystem/lib/aiohttp/__init__.py
+++ b/internal_filesystem/lib/aiohttp/__init__.py
@@ -107,11 +107,11 @@ class _RequestContextManager:
 
 
 class ClientSession:
-    def __init__(self, base_url="", headers={}, version=HttpVersion10):
+    def __init__(self, base_url="", headers=None, version=HttpVersion10):
         self._reader = None
         self._base_url = base_url
         self._base_headers = {"Connection": "close", "User-Agent": "compat"}
-        self._base_headers.update(**headers)
+        self._base_headers.update(**(headers or {}))
         self._http_version = version
 
     async def __aenter__(self):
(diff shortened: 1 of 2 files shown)


ThomasFarstrike:
Looks great, thank you so much!

PR #182, opened 2026-06-29
fix(net): resume downloads after a mid-stream connection drop

jnuyens:
Follow-up from the networking shortlist (the `download_manager.py:285` item).

### Problem
The per-chunk retry loop re-read the **same** aiohttp reader after a failure.
Once the underlying TCP/TLS connection dropped — common on flaky badge WiFi —
that reader was dead, so all 3 retries failed and the entire download aborted
with `OSError(-110)`. The resume infrastructure (`Range`/`Content-Range`,
`get_resume_position`) already existed but was never used on retry, so a blip
near the end of a large album/`.mpk` download threw away all progress.

### Fix
Restructure the read path into a **connection-retry loop**:
- On a read error, close the connection and re-issue the request with
  `Range: bytes=<partial_size>-`, continuing to write to the same `fd`.
- A server that supports ranges replies `206` → the download resumes from where
  it dropped. One that replies `200` (no range support) raises — same outcome as
  before, so no regression for those servers.
- Write/callback errors propagate immediately (they aren't connection problems,
  so they must not trigger a reconnect).
- Per-connection setup (status check, `Content-Range`/`Content-Length` parse,
  `fd` open) runs once on the first connection; reconnects skip it.
- Reconnect budget reuses `_MAX_RETRIES` (3).

### Testing
New `TestDownloadResumeOnConnectionDrop` (fake aiohttp via `sys.modules`, same
style as `TestRedactedExceptionPath`):
- `test_resumes_after_midstream_drop`: first connection drops at ~5 KB, the
  reconnect carries a `Range:` header, and the reassembled file equals the full
  12 KB payload (1 initial + 1 resume connection).
- `test_clean_download_uses_single_connection`: no drop → one connection, no
  `Range`, full payload.

Full suite: **36 tests OK** (34 existing + 2 new). `mpy-cross` compiles; ASCII-only.

Note: the maintainer's paired "fd not closed via try/finally" was verified to be
a false positive — `fd` is already closed on every path — so that part is a no-op.

### DIFF
--- a/internal_filesystem/lib/mpos/net/download_manager.py
+++ b/internal_filesystem/lib/mpos/net/download_manager.py
@@ -196,148 +196,158 @@ class DownloadManager:
         try:
             headers = cls._merge_headers(headers)
             
-            async with session.get(url, headers=headers, ssl=sslctx, timeout=_CHUNK_TIMEOUT_SECONDS) as response:
-                if response.status < 200 or response.status >= 400:
-                    logger.error("HTTP error %s", response.status)
-                    raise RuntimeError(f"HTTP {response.status}")
-                
-                # Figure out total size and starting offset (for resume support)
-                # When redacting, suppress the headers dump entirely — response
-                # headers can include `set-cookie`, `cf-ray` and other tokens
-                # that correlate to the request's secret-bearing URL.
-                if redact_url:
-                    if __debug__: logger.debug("Response headers: <redacted>")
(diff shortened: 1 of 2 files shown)


ThomasFarstrike:
Ah, sounds like a good fix! At one point this used to work but it probably got broken somewhere along the way... thanks for revamping it!

PR #184, opened 2026-07-03
Fix/wasm touch scroll

DrSkunk:
This potentially fixes the scroll issue on mobile devices.


ThomasFarstrike:
Great! The main build fails due to an unrelated issue so just ignore that, I'm fixing it and will then test this!

PR #187, opened 2026-07-09
Add setting to mute bootup sound

DrSkunk:
Add a setting to the menu to mute the bootup sound.


ThomasFarstrike:
I just hope this doesn't make the boot slower than it already is...

PR #188, opened 2026-07-09
remove mac intel build

DrSkunk:
Since macOS with Intel chips are from 2019 and the build stage takes a long while and is expensive in CI/CD tokens, I propose to remove the autobuilding for macOS with Intel chips. It is still possible to manually clone and build yourself.


ThomasFarstrike:
I didn't even know CI/CD tokens were a thing ;-D

As someone still using my Dell XPS Linux laptop from 2012, 2019 seems pretty recent... but any excuse to reduce support for a closed source system is good for me! ;-)

ThomasFarstrike:
> It is still possible to manually clone and build yourself

Theoretically yes, but in practice, disabling the autobuild will result in support being de-facto dropped because of "build rot" - subtle build issues not getting caught until someone tries it after a long while and finding out it no longer works...

Let me check with a few known macOS users in the chat to see if they are all okay with this.

PR #189, opened 2026-07-09
feat(net): cache async DNS resolutions to reuse across requests

jnuyens:
## What

`async_dns.getaddrinfo_async` now caches successful resolutions per `(host, port, proto, socktype)` with a 5-minute TTL, so repeated lookups for the same host reuse the result instead of spawning a resolver worker thread each time.

## Why

`getaddrinfo_async` spawns a `_thread` worker per lookup (correctly, to keep the loop alive during blocking DNS). But the ESP32-S3 thread pool is small, so any workload that resolves the same host repeatedly can exhaust it, after which new lookups fail with `OSError: can't create thread` and downstream requests fail.

This showed up in the wild on a Fri3d badge: streaming an album (a fetch per fragment) alongside a background app that constantly retried unreachable relays drained the pool, and both then hit `can't create thread`. The fetches and the retries were all re-resolving the same handful of hosts every time. Caching the resolution removes those redundant threads entirely, so retries and repeat downloads become pure-async.

## Details

- Cache check runs **before** the `sys.platform == "linux"` branch, so it applies on both desktop and ESP32 (and is exercised by the desktop tests).
- Only **successful** results are cached; failures re-resolve (so a transient error is not sticky).
- `clear_dns_cache()` is exposed for network changes / tests.
- 5-minute TTL keeps entries fresh enough to follow DNS changes.

## Tests

`tests/test_async_dns.py`: `setUp`/`tearDown` clear the cache; two new tests verify a repeated lookup hits the cache (no re-resolve) and that a failed lookup is not cached. All pass on the desktop build (`./tests/unittest.sh tests/test_async_dns.py`).

Follow-up to the off-loop DNS work in `async_dns`.

### DIFF
--- a/internal_filesystem/lib/mpos/net/async_dns.py
+++ b/internal_filesystem/lib/mpos/net/async_dns.py
@@ -25,6 +25,7 @@ Usage:
 import socket
 import _thread
 import sys
+import time
 from mpos.task_manager import TaskManager
 
 # Module-level reference to the getaddrinfo implementation.  Tests replace this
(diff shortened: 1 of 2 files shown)


ThomasFarstrike:
Ah that's nice, a DNS cache was needed!

PR #190, opened 2026-07-09
feat(tfl): run on WASM via async networking, and stream via an SD cache

jnuyens:
## What

Two changes to The Free Lantern Player, in `tfl_player.py`:

1. **Run on the WebAssembly build.** Migrate off `_thread` + `urequests` to `asyncio` + `DownloadManager`. The browser/WASM build has no raw sockets and no usable `_thread`, so the old blocking design couldn't run there. The worker, both play loops (`StreamPlayer`/`SDPlayer`), and the prefetcher are now async tasks; the three `requests.get` sites become `await DownloadManager.download_url(...)`; and every `time.sleep_ms` becomes `await TaskManager.sleep_ms`, so the app never blocks the LVGL event loop. Modelled on the app-store / OS-update apps, which already work on WASM.

2. **Stream via an SD cache when present.** Streaming previously required ~2 MiB free *internal* storage for its fragment cache and ignored an inserted SD card. It now caches to the SD card when one is present (falling back to internal), with a small free-space check on the chosen filesystem. Streaming works on badges with little free internal storage.

Behaviour is otherwise preserved: 45s fragment streaming, the 2-fragment prefetch buffer, next/prev-song jumps, the buffering banner, SD full-album download + verify, volume, and the About dialog.

## Testing

- Runs natively on a **Fri3d 2024 badge (OS 0.14.2)**: streams fragments over HTTPS and plays them, with the SD cache path exercised (no "needs 2 MiB" block).
- Desktop simulator: index load + UI + async download path.

## Notes

One documented trade-off: `get_remote_content_length` is removed because `DownloadManager` exposes no HEAD/Content-Length probe to callers, so the SD 10-day re-verify now detects added/removed fragments rather than byte-size changes of same-named ones (noted inline in the source).

Serving the fragments to an in-browser WASM client also needs the server to send `Access-Control-Allow-Origin` and use `https://` — those are server-side and out of scope for this app change.

### DIFF
--- a/internal_filesystem/apps/com.micropythonos.thefreelanternplayer/assets/tfl_player.py
+++ b/internal_filesystem/apps/com.micropythonos.thefreelanternplayer/assets/tfl_player.py
@@ -3,7 +3,7 @@
 #
 # The Free Lantern Player for MicroPythonOS (ESP32-S3).
 # - Streams 45s WAV fragments from index.html (Song Title<B>Track No<B>Album<B>frag1<B>frag2...)
-# - Keeps a 2-fragment buffer using a background _thread prefetcher
+# - Keeps a 2-fragment buffer using a background prefetch task
 # - Saves playback state every 10s so it can resume after restart
 # - Next/Prev jumps to next/prev SONG (not fragment)
 # - Shows "Buffering..." when waiting for the next fragment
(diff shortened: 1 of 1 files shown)


ThomasFarstrike:
Awesome!

PR #192, opened 2026-07-11
fix get_ipv4_netmask to get the netmask (not the address)

lebeno:
## Problem
The **IPv4 Netmask** in the About app showed the same value as **IPv4 address**.

## Reason
As documented in https://docs.micropython.org/en/latest/library/network.html#network.AbstractNIC.ipconfig, addr4 (e.g. 192.168.0.4/24) obtains the current IPv4 address and network mask as (ip, subnet)-tuple.

## Solution
* Add **tuple_index** (defaults to `0`) to the **_get_ipv4_value** method in wifi_service.py.
* Change the **get_ipv4_netmask** method to get the second element from the addr4 tuple (index=1).

## Testing
Compiled and successfully tested on Fri3d Camp 2026 badge hardware.

### DIFF
--- a/internal_filesystem/lib/mpos/net/wifi_service.py
+++ b/internal_filesystem/lib/mpos/net/wifi_service.py
@@ -513,7 +513,7 @@ class WifiService:
 
 
     @staticmethod
-    def _get_ipv4_value(network_module, ap_index, sta_key, desktop_value, label):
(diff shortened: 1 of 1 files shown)


ThomasFarstrike:
Good one, thanks and congratulations on your first contribution!

PR #193, opened 2026-07-11
 fix dead _webterm stdout mirror

DrSkunk:
The deployed web build's host bridge never received REPL output. `unix_mphal.c.patch` silently failed.
This should fix it so I can integrate the web version in the Fri3d IDE.


DrSkunk:
Also added virtual buttons, joystick and neopixels.
Screenshot 2026-07-11 at 16 03 55

ThomasFarstrike:
Awesome! Beautiful!

Could it be that the "Start" button is missing?

I would propose to put the "Menu" and "Start" buttons in the middle, like on the physical device, no?

PR #197, opened 2026-07-12
Re-export mpos.lights from the top-level mpos package

### LINKED ISSUE #196: mpos.lights / LightsManager not re-exported from the top-level mpos module
## What's wrong

[architecture/frameworks/](https://docs.micropythonos.com/architecture/frameworks/)
is explicit that apps should only ever do `from mpos import X` (or
`import mpos; mpos.X`) and that submodule imports like `from mpos.ui import
DisplayMetrics` should be avoided:

> If you find yourself needing to import from a submodule, or if you would
> like a new framework to be created, please create a GitHub issue
> describing your use case.

The onboard NeoPixel LED framework (`LightsManager`, in `mpos.lights`) isn't
in the documented "Available Frameworks" list and isn't re-exported from
`mpos/__init__.py`, so there's currently no way to use it without violating
the documented import convention:

```python
# what the docs say to avoid, but the only way that currently works:
import mpos.lights as LightsManager
```

## Use case

Using the 5 onboard NeoPixel LEDs below the screen as a life-counter
indicator in a game
([bomberboy](https://github.com/tjorim/bomberboy-mpos)'s `led_indicator.py`).

## Suggested fix

Add `LightsManager` to `mpos/__init__.py`'s re-exports (and to the
Available Frameworks list in the frameworks doc), matching every other
framework's pattern (`from mpos import LightsManager`).

tjorim:
Every other framework follows the documented "from mpos import X" import convention (architecture/frameworks/), enforced via re-exports in mpos/__init__.py. mpos.lights (LightsManager) was the one exception -- not in the Available Frameworks list and not re-exported, forcing the submodule import the docs explicitly say to avoid. Even the Fri3d 2026 board's own init code (board/fri3d_2026.py) has to reach for `import mpos.lights as LightsManager` as a result.

Adds it to the same "from . import X" + __all__ pattern already used for ui/net/audio/etc., matching the plain-module submodule-exposure style those already use (lights.py exposes module-level functions, not a class, so this mirrors that rather than the class-based singleton pattern used by AudioManager/SensorManager/etc.).

Closes #196

### DIFF
--- a/internal_filesystem/lib/mpos/__init__.py
+++ b/internal_filesystem/lib/mpos/__init__.py
@@ -73,6 +73,7 @@ from . import sensor_manager
 from . import camera_manager
 from . import sdcard
 from . import audio
+from . import lights
 
 __all__ = (
     # Core framework
@@ -118,7 +119,7 @@ __all__ = (
     "retry_action_until", "wait_for_focus",
     # Submodules
     "ui", "shared_preferences", "net", "content", "time", "sensor_manager",
-    "camera_manager", "sdcard", "audio",
+    "camera_manager", "sdcard", "audio", "lights",
     # Timezone utilities
     "TimeZone"
 )


ThomasFarstrike:
Ah good catch, yeah this API grew organically over time so anything you can spot that can be improved, sure!

Ultimately, the goal is to reach an "API Level 1" just like on Android, so that apps have a guarantee that they will remain supported, through MPOS updates, as long as they follow API Level N.

BadgeHub.eu would then also add the category mpos_api_1 (or something more human readable...) of course.

But for now, I have the feeling that the API still has some of these little defects that make it not ready for Level 1...

ThomasFarstrike:
Note that I reworked the LightsManager API to change "import mpos.lights as LightsManager" to the MPOS standard "from mpos import LightsManager", see https://docs.micropythonos.com/frameworks/lights-manager/

Issue #191, opened 2026-07-09
Nostr relay service retries unreachable relays with no backoff, exhausting the ESP32 thread pool and breaking other apps' networking


jnuyens:
## Summary

The background Nostr service (`com.micropythonos.nostr`, `nostr_boot_service`) retries unreachable relays with no backoff. On a Fri3d 2024 badge this drained the ESP32-S3 thread pool and broke **other apps' networking** — The Free Lantern Player could not fetch fragments during playback and the audio broke up badly.

## What happens

The service connects to three relays that are currently failing:

```
ERROR:nostr.relay:relay error wss://relay.damus.io: b'HTTP/1.1 522 '
ERROR:uaiowebsocket:_async_main's await self._connect_and_run() for wss://relay.primal.net got exception: -113
ERROR:uaiowebsocket:_async_main's await self._connect_and_run() for wss://relay.0xchat.com got exception: -113
```

It reconnects immediately and continuously. Each connection attempt resolves the relay host via `async_dns.getaddrinfo_async`, which spawns a `_thread` per lookup. With no backoff, the retries spawn resolver threads faster than the small ESP32-S3 pool allows, and it exhausts:

```
WARNING:notification_manager:Failed to play notification sound: can't create thread
...
File "mpos/net/async_dns.py", line 106, in getaddrinfo_async
OSError: can't create thread
TFLPlayer prefetch error (retrying): can't create thread
ERROR:nostr.relay:relay error wss://relay.damus.io: can't create thread
```

At that point `_thread.start_new_thread` fails everywhere on the device (I confirmed a bare `_thread.start_new_thread` from the REPL also raised `can't create thread`). Disabling the Nostr boot service made the pool healthy again and TFL streamed smoothly.

## Root cause / two fixes

1. **DNS re-resolution per attempt** (the thread-exhaustion trigger). A DNS cache in `async_dns` removes this: retries reuse the cached relay IP instead of spawning a resolver thread each time. Submitted as #189 — it makes the retries pure-async and stops the thread starvation for Nostr and any other repeat-lookup workload.

2. **No relay backoff** (the remaining waste). Even with the DNS cache, the service still hammers unreachable relays as fast as it can, wasting CPU and network. It should back off on failed connections (e.g. exponential backoff per relay, capped), so a set of down relays does not spin the event loop.

Fix #1 (via #189) stops it from breaking other apps; fix #2 is the proper behavior for the service itself.

## Environment

[...]

PR #201, opened 2026-07-13
fix(nostr): exponential reconnect backoff so unreachable relays stop exhausting the thread pool (#191)

### LINKED ISSUE #1: Just wanted to let you know.
I thought this would be something you would be interested in. I am currently working on exposing FreeRTOS to the Python interpreter in MicroPython. This would allow for you to have real threads and also provide access to real parallel processing by being able to set "threads" to specific cores on the ESP32 modules. You would need to disable the GIL (Global Interpreter Lock) in MicroPython for it to work so care has to be taken to properly control access to data objects from different "threads".

Thought you might be interested in something like that considering the scope of this project.

### LINKED ISSUE #2: Build process or ESP32 is broken
The build process for the ESP32 as described here [https://docs.micropythonos.com/building/esp32/](https://docs.micropythonos.com/building/esp32/) fails.
It looks like this is due to hard coded paths.

jnuyens:
Fix #2 for #191. Builds on the reconnect rework already merged to `main`.

## What #191 asked for
The Nostr relay service retried unreachable relays with no backoff. Each attempt resolves DNS and spawns a `_thread`, and the ESP32-S3 pool is tiny, so hammering a dead relay starved other apps' networking (this is what broke audio playback in The Free Lantern Player). The DNS cache (#189) was fix #1; this is fix #2.

## What was still wrong on `main`
`main` already normalises `reconnect_interval` and gates on `_network_available()`, which handles the "Wi-Fi fully down" case. But two gaps remained when Wi-Fi is up and a specific relay is unreachable or flapping:

1. **Clean-close tight loop.** When a relay accepts the socket then drops it, `_connect_and_run()` returns without raising, so the `except` branch (with its `sleep`) never runs. The `while self.running:` loop reconnects with **zero delay** and hammers.
2. **No backoff.** The exception path waits a fixed `reconnect_interval` (3s) forever, so a permanently dead relay keeps retrying every 3s.

## The fix
Exponential backoff in `_async_main`, seeded from `reconnect_interval`:

- Wait before **every** reconnect (not just after an exception), which closes the clean-close zero-delay loop.
- `connected_ok` tracks whether the just-ended session was actually live: reset the delay to the seed interval after a live close, otherwise double it toward a 300s ceiling. An unreachable relay backs off (3s -> 6 -> 12 -> ... -> 300) instead of spinning; a healthy relay that blips recovers fast.
- No relay is abandoned, so one that recovers reconnects on its own.

The backoff math is a pure helper (`_next_backoff`) so it is unit-testable without a network. `reconnect_interval <= 0` still disables reconnect, and `_network_available()` gating is unchanged.

Scope is exactly the Nostr relay service: `WebSocketApp` is only used by `nostr/relay.py` (nostr app + NWC both go through it).

## Testing
- New `tests/test_uaiowebsocket_backoff.py`: 5/5 pass on the desktop build (`./tests/unittest.sh tests/test_uaiowebsocket_backoff.py`), network-free. Covers geometric growth, the 300s cap, reset-on-live-connection, and a custom seed interval as the floor.
- `mpy-cross` compiles clean; ASCII-only, MicroPython-compatible.

### DIFF
--- a/internal_filesystem/lib/uaiowebsocket.py
+++ b/internal_filesystem/lib/uaiowebsocket.py
@@ -54,6 +54,16 @@ class WebSocketConnectionClosedException(WebSocketException):
 class WebSocketTimeoutException(WebSocketException):
     pass
 
+# Reconnect backoff ceiling (seconds): a failing relay must not be retried in a
+# tight loop (each attempt spawns a DNS _thread; the ESP32 pool is tiny) (#191).
+_RECONNECT_MAX_S = 300
+
+def _next_backoff(current, connected_ok, min_s, max_s=_RECONNECT_MAX_S):
+    """Reset to min after a live connection closed, else grow toward max."""
+    if connected_ok:
+        return min_s
+    return min(current * 2, max_s)
+
 # Queue for callback dispatching
 _callback_queue = ucollections.deque((), 100)  # Empty tuple, maxlen=100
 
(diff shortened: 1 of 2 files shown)


ThomasFarstrike:
Oh that's very cool! Thanks for adding a unit test too - the more, the merrier! ;-)

PR #204, opened 2026-07-16
LightsManager fix for Freenove ESP32-S3

scott-macg:
Fixes a TypeError during boot on the Freenove ESP32-S3 board (AKA Hosyond ES3C28P) where LightsManager.init() was called with the deprecated num_leds keyword argument. This was preventing MPOS from initializing.


ThomasFarstrike:
Ow, I'm sorry for breaking it, thanks for the fix!

Rohansi:
Thanks! This fixed it for me too

PR #216, opened 2026-07-26
Add publish-esp32s3-badgehub.yml GitHub Action

francisduvivier:
This is a github action to automatically publish the esp32s3 bin file to BadgeHub when a MicroPythonOS release is created.

I chose to only publish esp32s3 and strip version in filename to keep things simple, but this can be discussed ofcourse.

Tested on my fork:
- Github action run: https://github.com/francisduvivier/MicroPythonOS_fork/actions/runs/30196717925
- Test release: https://github.com/francisduvivier/MicroPythonOS_fork/releases/tag/0.16.0
- Test app: https://badgehub.eu/page/project/test_admin on

To make this work on the MicroPythonOS repo, this still requires:
- An app called `com.micropythonos.esp32s3` to be created on badgehub.eu
  - On badgehub, The name, Icon, description, category, ... can optionally be configured, this metadata will **not** be overwritten by this action.
  - Also, ideally, the file should be set as the main executable manually after the first upload.
- A token to be created for that project and set in GitHub repo secrets as `BADGEHUB_PROJECT_TOKEN`

### DIFF
--- a/.github/workflows/publish-esp32s3-badgehub.yml
+++ b/.github/workflows/publish-esp32s3-badgehub.yml
@@ -1,0 +1,101 @@
+name: Publish MicroPythonOS for ESP32S3 to BadgeHub
+
+on:
+  release:
+    types:
+      - published
+
+env:
+  BADGEHUB_API_URL: ${{ vars.BADGEHUB_API_URL || 'https://badgehub.eu/api/v3' }}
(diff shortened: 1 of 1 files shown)


francisduvivier:
Closing since we actually want to have this in the fri3d repo instead.

PR #219, opened 2026-07-28
build_mpos.sh: restore native/viper decorators after desktop builds

bitcoin3us:
## Summary

`disable_native_viper()` comments out `@micropython.native` / `@micropython.viper` across `internal_filesystem/` for the unix, macOS, and web targets (the decorators must not be frozen), but nothing ever reverses the edit. In CI that's harmless — the checkout is throwaway, and the esp32 builds run first so they keep the optimized decorators. For local development, every desktop/web build leaves the touched files (`draw.py`, `appstore/blurhash.py`, `mpos/audio/adpcm_ima.py`, `mpos/audio/stream_wav.py`, ...) modified in git, which is easy to mistake for real work — or worse, to accidentally commit or discard alongside real work.

This adds `restore_native_viper()` as the exact inverse (uncomment the decorators the disable step commented out) and invokes it via an **EXIT trap**, so the tree is restored on build success, build failure, and CTRL-C alike. The trap is only installed after `disable_native_viper` has actually run, and the restore happens after freezing, so the frozen bytecode is unaffected.

## Test plan

- [x] Round-trip test: a file containing both `@micropython.native` (column 0) and an indented `@staticmethod`-stacked `@micropython.viper` is byte-identical after disable + restore
- [x] `scripts/build_mpos.sh macOS` completes and `git status` shows no modified `internal_filesystem/` files afterwards

## Merge checklist

- CHANGELOG.md updated under "Future release" (Development section)
- No settings/app changes; no test changes applicable (shell build script)

🤖 Generated with [Claude Code](https://claude.com/claude-code), running with bypass permissions enabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ OS:
 - sdl_keyboard: fix CTRL-C and CTRL-V on textarea if set_text_selection(True)
 
 Development:
+- build_mpos.sh: restore @micropython.native/@micropython.viper decorators after desktop/web builds (via EXIT trap) so local builds no longer leave the working tree modified
 - Add Python-level line coverage via sys.settrace (mpcov build variant)
 - Add `--coverage` flag to test_runner.py for collecting per-file line coverage
 - Add `make build-mpos-unix-coverage` target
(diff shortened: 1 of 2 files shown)


DrSkunk:
Thanks! that was bugging me too.

PR #227, opened 2026-08-04
mpos_controller: accept CRLF sentinel endings in read_until

fdb:
## The bug

`AIOREPLClient.exec()` brackets device output with sentinels and calls `read_until(END_MARKER + b"\n")` to know when the result is complete. That suffix check assumes a bare LF — but the newline convention differs per transport:

- **Desktop (PTY backend)**: the controller clears `OPOST` on the pty, so the device's `\n` arrives untranslated. The sentinel matches immediately.
- **Serial (real device)**: the REPL emits CRLF. The buffer ends with `~~~MPOS_END~~~\r\n`, which does **not** end with `~~~MPOS_END~~~\n`. The match never succeeds.

What makes this hard to notice: after the 30s timeout, `read_until` returns everything it read anyway, and `exec()` then extracts the output between sentinels with `rfind` — successfully. So every serial `exec`/`eval` returns the **correct answer, 30 seconds late, with no error**. A single CLI invocation stacks four of them (three display-resolution `eval`s in `SerialBackend.start()` plus the requested `exec`).

## Measured on a Fri3d 2026 badge

| | before | after |
|---|---|---|
| `exec 'print("hello")'` over serial | 2m04s | 6.6s |
| no-change app deploy (4 controller calls) | 6m23s | 34s |

## The fix

When the expected ending is LF-terminated, also accept its CRLF twin — `bytes.endswith()` takes a tuple, so the hot loop gains no extra work. Endings that already are CRLF, or that don't end in a newline (like the `b">>> "` prompt waits), are unaffected. The desktop PTY path matches the first tuple element exactly as before.


ThomasFarstrike:
Note that it's possible some tests fail if you're not on the latest commit because I just reworked the CI/release workflows.

fdb:
The cancelled `test` job (macOS Intel) looks like the known CI instability rather than this change: the same workflow was cancelled or failed on every run today, including on `main` itself (e.g. the "ci: bump all CI workflow timeouts to 120 min" and "Prepare for release" commits). The job hung ~30 minutes after its last test output and was killed with orphan Python processes — this PR's run raced under the pre-bump timeout.

Rebased onto current `main` (af4475e) to pick up the 120-min timeouts and rerun. The change itself only touches host-side sentinel matching in `read_until`; `lint` and both macOS builds passed on the previous run.

fdb:
Would be nice if you could take a look at this, since it makes badge deploys during development *much* faster.

ThomasFarstrike:
Of course! Just had to push out the release 0.16.1 first :-)

ThomasFarstrike:
This is awesome!

Just one remark; there's a tests/cpython_mpos_controller.py but no new test case was added to this PR here for CRLF sentinel matching. Given the exec() fallback (rfind) already recovers correctly, a unit test for read_until itself would catch future regressions without needing device hardware...

We could add the test later, but if you care to give it a go, that would make the most sense... Please let me know!

fdb:
Hey, just added tests for this, covering four cases: LF output (desktop PTY), CRLF output (the serial regression), the bare >>>  prompt ending, and the timeout-returns-data fallback that exec()'s rfind relies on.

One note on the CRLF case: since the old code still returned the correct bytes after the timeout, the content check alone passes against it. So we've added an elapsed-time check — if we see that the test takes 10s, that's a regression, and the test turns red.

ThomasFarstrike:
Sorry to bother you again but after merging https://github.com/MicroPythonOS/MicroPythonOS/pull/226 there's 3 merge conflicts... would you mind...? I could guestimate it myself but I'd rather have your _informed_ resolution :-)

I read something about GitHub adding support for patch sets or something yesterday so that might help with these types of situations!

ThomasFarstrike:
Thank you!

Issue #229, opened 2026-08-05
LoRa driver unification: replace custom SX1262 driver with upstream micropython-lib + adapter + framework watchdog


ThomasFarstrike:
## Background

We currently carry two separate, divergent SX1262 LoRa drivers:

| Board | Driver | Lines | SPI API |
|-------|--------|-------|---------|
| `fri3d_2026` | `lib/drivers/lora/sx1262.py` | 2426 | `SPI.Device` (lvgl_micropython Bus/Device split) |
| `lilygo_t_watch_s3_plus` | `lib/drivers/lora/micropySX126X/` | 1695 total | Raw `SPI` pins, with SoftSPI fallback |

Both are forks of older community drivers (`micropySX126X` by ehong-tl), diverged for board-specific reasons (SPI API differences, instability workarounds, blocking/non-blocking patterns). They share no code and drift independently.

The ["sword of Damocles"](https://github.com/MicroPythonOS/MicroPythonOS/pull/222#issuecomment-5133313236) is that neither tracks upstream, and both have accumulated board-specific hacks that make them fragile and hard to maintain. PR #222 patched one symptom (CS held low across BUSY wait) but did not address the underlying duplication.

**The goal**: replace both with the [official upstream `micropython-lib` lora driver](https://github.com/micropython/micropython-lib/tree/master/micropython/lora) by Angus Gratton, using thin adapter layers + `.patch` files (only if unavoidable) so we can easily follow upstream releases.

## Scope

This issue covers the **fri3d_2026** driver replacement + framework additions. The **lilygo_t_watch_s3_plus** driver is deferred to a follow-up PR to keep scope tight.

---

## Phase 1: Copy upstream files (verbatim)

Copy 4 files from `~/sources/micropython-lib/micropython/lora/` into `internal_filesystem/lib/lora/`:

| Source | Destination | Lines |
|--------|------------|-------|
| `lora/lora/modem.py` | `lib/lora/modem.py` | 474 |
| `lora-sx126x/lora/sx126x.py` | `lib/lora/sx126x.py` | 899 |
| `lora-sync/lora/sync_modem.py` | `lib/lora/sync_modem.py` | 86 |
| `lora/lora/__init__.py` | `lib/lora/__init__.py` | ~20 |

- `__init__.py` trimmed to import only `sx126x` (drop `sx127x`, `stm32wl5`).
- These files are **never modified directly**. Any needed changes go in the adapter or as `.patch` files.

[...]

Issue #224, opened 2026-08-01
fri3d_2026: Expander.config register write doesn't take effect at runtime (CH32 fw v2.0.1)


steemandavid:
## Summary

On a Fri3d Camp 2026 badge (CH32 coprocessor firmware **v2.0.1**), writing to the
`Expander.config` register (`drivers/fri3d/expander.py`, I2C register `0x16`) at
runtime — after the board has finished booting — appears to have **no effect**: reading
`config` back immediately after the write (or 100ms later) shows the same value as
before the write, every time. A different register on the same expander,
`lcd_brightness`, updates correctly and immediately via the same write path
(`Device._write` → `i2c.writeto_mem`), so this looks specific to `config` rather than a
general I2C problem.

This matters because `config` bit 4 is the only way to hardware-reset the LoRa
(SX1262) chip on this board — its reset pin isn't wired to the ESP32-S3 directly, only
to the CH32 (see `board/fri3d_2026.py`'s own boot-time use of `expander.config = 0x01`
/ `0x13`, and [lucid-void/fri3d-meshcore#7](https://github.com/lucid-void/fri3d-meshcore/issues/7),
which relies on toggling this same register at runtime to recover a wedged radio).
If runtime writes to `config` are silently ignored post-boot, that recovery path
doesn't actually work, which seems worth knowing either way.

## Steps to reproduce

From a MicroPython REPL on a running fri3d_2026 badge (any app in the foreground):

```python
import mpos
exp = mpos.io_expander

print('config before:', exp.config)       # e.g. (True, False, False, True, True)
exp.config = 0x03                          # would assert LoRa reset per the bit layout
print('config after write 0x03:', exp.config)   # unchanged, still the pre-write value

exp.config = 0x13
print('config after write 0x13:', exp.config)   # unchanged again

# control: a different expander register, written the same way, DOES take effect
print('brightness before:', exp.lcd_brightness)
exp.lcd_brightness = 30
print('brightness after:', exp.lcd_brightness)  # correctly reads back 30
```

Also reproduced with a raw register write, bypassing the `config` property entirely:

```python
print(exp._read('B', 0x16, 1))          # e.g. (19,)
exp._write(0x16, bytes([0x03]))
print(exp._read('B', 0x16, 1))          # still (19,) -- unchanged
```

## Expected

`config` reads back the value that was just written (as `lcd_brightness` does), or —
if the reset bit is intentionally a self-clearing "pulse" abstraction rather than a
[...]

PR #231, opened 2026-08-05
lora: replace custom fri3d_2026 SX1262 driver with upstream micropython-lib driver + adapter

ThomasFarstrike:
Replace the 2426-line forked drivers/lora/sx1262.py with the official upstream micropython-lib lora driver (4 files, ~1450 lines verbatim) plus thin adapter layers that encapsulate all board-specific quirks.

New files:
  lib/lora/                 - upstream lora package (modem, sx126x, sync_modem, __init__)
  mpos/lora_spi_adapter.py  - wraps SPI.Device as standard machine.SPI
  mpos/lora_adapter.py      - wraps upstream SX1262 in existing caller API

Enhanced:
  mpos/lora_manager.py      - acquire/release lock, CH32 reset_chip(),
                               optional watchdog with auto-recovery

Updated callers:
  board/fri3d_2026.py       - import from mpos.lora_adapter
  apps/lora_chat/lora_chat.py - import + acquire/release + watchdog

Deleted:
  drivers/lora/sx1262.py    - replaced by upstream + adapter

Single upstream deviation: from machine import Pin guarded with try/except in sx126x.py (Unix desktop port has no machine.Pin). No other changes to upstream files.

LoRaManager now provides:
  acquire(app_name)/release(app_name)  - exclusive radio lock
  reset_chip()                         - CH32 IO expander reset
  start_watchdog()/stop_watchdog()     - auto-recovery from SPI wedges
  is_healthy()                         - status probe

micropySX126X/ (lilygo watcher driver) left untouched for a follow-up PR.

Refs: PR #222, issue #229


ThomasFarstrike:
Integrating the upstream LoRa driver itself was relatively easy but so far I was unable to make the whole perfectly stable.

Some challenges:
- The SPI bus being shared with the display (which is doing lots of SPI writes, especially during scrolling and animations) and also with the SD card, means there's no way to talk to the LoRa chip without blocking the display and the SD card.
- MicroPython ignoring IRQ (interrupts) when the system is busy (like when updating the display etc)
- There was also a bug in the lvgl_micropython SPI driver which prevented DMA transfers, now fixed
- The LoRa chip seems to prefer byte-by-byte SPI communication instead of entire buffers, possibly due to some timing sensitivity?
- The LoRa chip is throwing errors, which are accidentally ignored when doing byte-by-byte transfers but not when doing entire buffers?
- Resetting the LoRa chip using the expander.config register had a bug (#224) which _should_ now be fixed, but working around that also cost time and muddied the water
- Testing the LoRa requires 3 devices (one for sending, one for receiving, one for observing) all having working software so it's quite tedious

So we made good progress, but since the final result isn't perfectly stable.

Meanwhile, people have built working apps on the existing driver, and Fri3d Camp 2026 is about to start.

So I think it's safer to play it safe by shipping only:
- the SPI DMA fix
- the SPI locking exposure

...while staying on the current LoRa driver for now.

Let's see how much the fixes help, how stable we can get the LoRa, and then revisit this.

PR #234, opened 2026-08-06
Fix permanent serial console death from mpremote/raw-REPL connections

bitcoin3us:
## Problem

On device (observed on MPOS 0.16.0, Waveshare ESP32-S3 Touch LCD 2), the aiorepl serial console permanently dies after a few `mpremote`/raw-REPL connection attempts. Each connection attempt sends Ctrl-C; KeyboardInterrupt tracebacks appear in `task_handler.py` (`_task_handler`/`_timer_cb`), and after several of these the aiorepl task itself is killed and never restarted. The LVGL UI keeps running, but the console is gone until a physical replug — making on-device app development a constant replug loop.

## Root causes

Three separate defects compound into the permanent console death:

1. **aiorepl exits for good on stdin EOF.** The MPOS EOF check in `aiorepl.task()` `return`ed when the host disconnected the serial port. Its `finally:` then left `kbd_intr(3)` armed, so every subsequent Ctrl-C from a connection attempt raised KeyboardInterrupt at an arbitrary VM checkpoint — visible as the `task_handler` tracebacks (its handlers only catch `Exception`, and `KeyboardInterrupt` is a `BaseException`).

2. **A stray KeyboardInterrupt tears down the entire asyncio loop.** MicroPython's `run_until_complete` only catches `(CancelledError, Exception)`, so a KeyboardInterrupt raised inside *any* task escapes `asyncio.run()` entirely, killing every task (console, wifi, services). The UI survives because the LVGL task handler runs on a hardware timer — exactly the observed "UI alive, console dead" state.

3. **Nothing ever restarts the REPL task** once it dies.

A fourth, related issue surfaced during on-device verification: **hosts connecting while the OS boots corrupt the boot itself.** mpremote's raw-REPL entry sends Ctrl-C during the ~10–40s boot window, which silently aborts the boot scripts (standard MicroPython behavior), leaving the device half-started at the friendly REPL — whose `>>> ` prompt masquerades as the OS console and makes `/lib` deploys appear not to take effect.

## Fix

- `internal_filesystem/lib/aiorepl.py` — on EOF, sleep 500 ms and keep reading instead of returning. No prompt re-print, so the desktop `/dev/null` case stays quiet (the original reason for the EOF `return`).
- `internal_filesystem/lib/mpos/task_manager.py` — `TaskManager.start()` now loops: a KeyboardInterrupt escaping `asyncio.run()` is logged and the loop re-entered. The global task queue survives the unwind, so all remaining tasks resume.
- `internal_filesystem/lib/mpos/task_manager.py` — new `TaskManager.create_supervised_task(coroutine_factory, restart_delay_ms=200)`: restarts the task if it dies from KeyboardInterrupt or Exception; clean return and cancellation end supervision.
- `internal_filesystem/lib/mpos/app/system_services.py` — `AIOReplService` runs the REPL via `create_supervised_task()`, so a dead console self-restarts.
- `internal_filesystem/lib/mpos/main.py` + `internal_filesystem/main.py` — disable `micropython.kbd_intr` while `mpos.main` boots and restore it on the REPL fallback paths (guarded with `hasattr` for the web port). A mid-boot mpremote connect now just waits for boot to finish instead of aborting it.

## Tests

New `tests/test_task_supervision.py` (runs via `./scripts/test_runner.py`, written failing-first):
- supervised task restarts after KeyboardInterrupt and after Exception
- clean return is not restarted; cancellation stops supervision
- `TaskManager.start()` resumes the loop after a KeyboardInterrupt escapes a task
- `aiorepl.task()` stays alive on persistent stdin EOF (StreamReader monkeypatched)

All existing tests importing the changed modules pass (websocket, webserver, osupdate, multi_connect, multi_websocket, graphical IMU calibration). `ruff check` clean on all changed files.

## On-device verification (Waveshare ESP32-S3 Touch LCD 2)

With the fix deployed to `/lib` and a healthy boot confirmed (LVGL screen active, `TaskManager.keep_running` True, supervised aiorepl answering):
- 8× mpremote connect → exec → disconnect cycles with Ctrl-C spam between each: console alive after every one
- `raise KeyboardInterrupt` injected inside the REPL task: supervisor restarted it
- 5× rapid open/close disconnect storm (previously fatal via the EOF path): console alive
- 19 mpremote-style Ctrl-C/raw-REPL byte bursts fired during the boot window: boot completed normally (`mpos.main` loaded, UI alive)

## Notes for reviewers

- With the boot guard active, a Ctrl-C can no longer break into a *hung* boot; a crashed boot still falls back to the friendly REPL with the interrupt character restored. The escape hatches for a hung boot remain the reset button / esptool.
- Deliberate exits are preserved: Ctrl-D in aiorepl still shuts down asyncio, and `mpos.TaskManager.stop()` still returns to the sync REPL (supervision does not restart on clean return or cancellation).


bitcoin3us:
Merge checklist (https://docs.micropythonos.com/other/merge-checklist/):

- **CHANGELOG.md** — expanded under "Future release (next version)" → OS, with one entry per behavior change.
- **App version numbers** — N/A, no app is modified (changes are in `lib/mpos/` and `lib/aiorepl.py`).
- **Documentation** — MicroPythonOS/docs#8 documents `TaskManager.create_supervised_task()` and the boot-time Ctrl-C handling.
- **MAINTAINERS.md** — N/A, no new board.
- **Setting migration** — N/A, no Setting definitions change.
- **Unit tests** — new `tests/test_task_supervision.py` covers supervision restart on KeyboardInterrupt and on Exception, no restart on clean return, cancellation stopping supervision, loop resumption after a KeyboardInterrupt escapes a task, and aiorepl surviving persistent stdin EOF. Plus the on-device verification described above.
- **Non-functional changes separated** — the diff is functional only; no reformatting or unrelated cleanups.

ThomasFarstrike:
Interesting, thanks for creating a pull request!

I haven't noticed this myself, but I'm on a different OS and different virtual machine USB serial redirect etc, so that might be the difference.

3 potentially relevant things I've noticed and would like your opinion on:


1) I've noticed is that, to exit the mpremote.py shell, I have to press CTRL-A (to enter "raw REPL mode") and then press CTRL-X to exit.

That works, but then when I reconnect with mpremote.py, the device is still in "raw REPL" mode and doesn't respond to normal input, making it appear like a "serial console death", until I press CTRL-B to go back to "normal REPL mode" and then I can type regular commands again.

When I exit the REPL shell using CTRL-] then I don't have that issue.

I just wanted to check and confirm that you're certain this "serial console death" you observe is not due to sending CTRL-A or having the device in raw REPL mode because of whatever reason.


2) Often times, when I connect to the REPL shell over serial the first time (using tio or mpremote.py or picocom or some other serial client), I notice that the serial has a bit of *input* on the REPL shell, and when I press enter (send a newline) then this input gets executed...

Where does that input come from? I don't know.

It doesn't happen every time, maybe it's just from noise from the serial rx line floating during reset... but I just wanted to ask if you could confirm that that's not the reason for the "serial console death". Perhaps the floating serial rx pin in your case generates CTRL-A or CTRL-C some other combo that causes the REPL to be in "raw REPL mode" or some other hanging mode...


3) Doing help() on the (aio)REPL shell explains these commands:

```
Control commands:
  CTRL-A        -- on a blank line, enter raw REPL mode
  CTRL-B        -- on a blank line, enter normal REPL mode
  CTRL-C        -- interrupt a running program
  CTRL-D        -- on a blank line, do a soft reset of the board
  CTRL-E        -- on a blank line, enter paste mode
```

Do I understand correctly that this pull request here would modify some of these, such as CTRL-C?

***

Anyway, looking forward to having a more stable system so hope this can help!

bitcoin3us:
Thanks for the careful review — points 1 and 2 sent me back to check, and they turned up a
second, more specific cause that the original commit didn't cover. I've pushed a follow-up.

**1) Is it just the device being left in raw REPL mode?**

I checked this directly, because it's the obvious explanation. It isn't the cause:

```
# deliberately leave the device in raw REPL, then disconnect
>>> write(b'\x01')  ->  b'raw REPL; CTRL-B to exit\r\n>'
# reconnect
$ mpremote connect /dev/cu.usbmodem1101 exec "import mpos; print(mpos.TaskManager.keep_running)"
True          # first attempt, no Ctrl-B needed
```

A device sitting in raw REPL is transparent to mpremote, because `raw_repl()` answers an
incoming Ctrl-A by re-printing the `raw REPL; CTRL-B to exit` banner, which is exactly what
`enter_raw_repl()` waits for.

Your observation is real though, and worth separating from this bug: with a plain terminal
(tio/picocom) a device left in raw mode *does* look dead until you press Ctrl-B. The
difference is the failure signature. In the state I'm reporting, mpremote fails with
`could not enter raw repl` — it sent Ctrl-A and got nothing back at all, because the REPL
task is no longer running. Recovery needs a power cycle; Ctrl-B does nothing.

**2) Stray input on the RX line**

This turns out to matter a lot, and it's what led me to the real cause. In
`aiorepl.task()`, a Ctrl-D received in the normal (non-raw) loop does:

```python
sys.stdout.write("\n")
# Shutdown asyncio.
asyncio.new_event_loop()
return
```

On MicroPythonOS that discards the entire task queue — the running app, TaskManager's main
task, and the REPL task itself — and then returns *normally*, so
`create_supervised_task()` treats it as "finished, nothing to restart". The console is gone
until a power cycle, while the LVGL UI carries on because it's driven by a hardware timer.
That combination — dead serial, live display — is exactly the symptom I kept hitting.

So yes: if a floating RX line produces a single 0x04, that alone kills the console
permanently. mpremote also sends Ctrl-D of its own accord (`enter_raw_repl(soft_reset=True)`
sends one right after the banner, and `mpremote soft-reset` sends one directly), so any
session that desyncs — an aborted transfer, a client killed mid-protocol, a reset race —
can leave one landing in the non-raw loop.

The follow-up commit makes that branch end the line like Ctrl-C instead of tearing down
asyncio, and adds `create_supervised_task(..., restart_on_return=True)` for the console so
that *any* quiet exit path is recovered from rather than only crashes.

Evidence that the test actually bites: with the fix reverted, the new
`test_ctrl_d_does_not_kill_the_asyncio_loop` doesn't merely fail — it crashes the test
runner itself, because the harness runs tests inside a booted MPOS and the Ctrl-D takes down
the harness's own loop. Same bug, same mechanism.

On-device after the fix: 5 raw Ctrl-Ds outside raw mode each returned a fresh prompt with
the app still running; 10 connect/Ctrl-C/Ctrl-D/disconnect cycles, 5 transfers killed
mid-flight, and 3 reset+immediate-reconnect races all left the console answering on the
first attempt.

**3) Does this change the documented control keys?**

Partly, and I should be upfront about which:

- The `help()` text you quoted is the **stock MicroPython REPL's**. aiorepl is a separate
  implementation and already differed here: its Ctrl-D was never a soft reset — it called
  `asyncio.new_event_loop()` and returned.
- **Ctrl-C is unchanged.** It still interrupts. The original commit only stops a
  KeyboardInterrupt escaping `run_until_complete` from taking every other task down with it
  (MicroPython's `run_until_complete` catches only `CancelledError` and `Exception`).
- **Ctrl-D is changed**, deliberately: in the non-raw loop it now ends the line instead of
  shutting asyncio down. Inside the raw-REPL protocol it is untouched, so mpremote's use of
  it is unaffected.

If you'd rather Ctrl-D kept a meaning closer to the documented one, I'm happy to change it
to a real `machine.soft_reset()`, or to gate the current behaviour behind a flag — say which
you prefer and I'll amend. My reasoning for "do nothing" is that on an OS where the whole
system lives in that event loop, a single stray byte shouldn't be able to take the device
down, and there's already a documented way to drop out deliberately
(`mpos.TaskManager.stop()`).

ThomasFarstrike:
It's hard to predict and test all different setups:
- serial clients (putty, picocom, mpremote.py, pyserial, WebSerial like https://fri3dcamp.github.io/Fri3d-IDE/)
- connection methods (USB routed to a virtual machine vs USB directly on the physical machine vs usbip/usbipd vs USB-TTL connected to the device's UART RX and TX pins)
- user behavior (like pressing CTRL-D out of habit to stop the aiorepl and get dropped into the normal repl)
- webrepl (+ different webbrowsers)

Also, this is very important "core" functionality, so we can't break it.

On the other hand, this "permanent serial console death" is a real issue you observed and I also had serial hangs (especially when doing automated testing, one after the other, on the device - it would hang after a few tests, regardless of doing machine.reset() between each test).

So I think the best way forward is to merge it, and for the next release (0.18.0) we need to really test the release candidate in as many scenario's as we can, to ensure no regressions with regards to the REPL.

Anyway, thanks for giving this so much thought!

ThomasFarstrike:
One more question @bitcoin3us - am I right to assume that this print on the serial at boot is still correct:

`511682449:WARNING:mpos.app.system_services:Starting very limited asyncio REPL task. To stop all asyncio tasks and go to real REPL, do: mpos.TaskManager.stop()`

...because the users might need a way to be able to stop the aiorepl and drop to the regular REPL...

Issue #6, opened 2025-10-15
Support for other devices (ESP32 Devkit V1, ESP32 C3 VevkitM-1)


GeorgeMoto:
Hi, I'm curious about one thing. 
In the Supported Hardware section, I only see two devices officially supported for this OS:
- ESP32 Microcontrollers
- Fri3d Camp 2024 Badge

Is it technically possible to install the OS on other microcontrollers? For example on ESP32 Devkit V1 or ESP32 C3 VevkitM-1. I have an idea to connect a touch display directly to the device.

I'd appreciate any information on this topic, thank you.

ThomasFarstrike:
Yes, sure! That requires porting it, which is always a bit technical and "depends on the exact hardware" kind of situation, but I don't see why not.

I just wrote this "Porting Guide" here so maybe that already answers your questions: https://docs.micropythonos.com/building/porting/

Of course, these devices are really cheap so depending on your use case, you might consider buying a Waveshare 2 Inch Touch costs for $15-20 and save you the trouble. They're fun to play with, regardless!

ThomasFarstrike:
Any update @GeorgeMoto ? I closed this issue as it's pretty generic, but feel free to send a pull request or create another, more specific issue to follow-up!

PR #237, opened 2026-08-09
Fix runtime-loaded apps with native/viper decorators failing on desktop builds without a native emitter

bitcoin3us:
Companion to MicroPythonOS/lvgl_micropython#6, which adds the actual `unix_native_decorator_fallback.patch`. This PR wires it into `scripts/build_mpos.sh` for the **unix/macOS** and **web** targets, plus a CHANGELOG entry. The apply is existence-guarded (same pattern as `imgfont_set_range.patch`), so this merges safely before or after #6; it becomes active once the lvgl_micropython submodule pointer includes the patch file.

## Problem
On architectures without a native emitter (unix port on **aarch64**/Apple Silicon; the **wasm** port), any runtime-loaded `.py` containing `@micropython.native` or `@micropython.viper` fails to compile: `SyntaxError: invalid micropython decorator`. `disable_native_viper` only covers **frozen** code; `apps/` and the on-disk `lib/` are compiled at runtime by the port.

On a fresh 0.16.1 macOS arm64 build:
- the Draw app fails to import (`draw.py` has `@micropython.native`)
- AppStore's `blurhash.py` fails (`@micropython.viper` x2)
- boot logs `failed to import service NostrBootService ... invalid micropython decorator` (desktop runs resolve `lib/nostr` from disk via symlink, and `nip44.py` has `@micropython.native`)

Reproduce: `./lvgl_micropython/build/lvgl_micropy_macOS -c "compile(open('internal_filesystem/apps/com.micropythonos.draw/draw.py').read(),'x','exec')"` on an arm64 Mac.

Historical note: before #219 (restore native/viper decorators after desktop builds) this was masked on dev machines, because a previous build would leave the decorators commented out in the working tree. #219 fixed that hygiene problem and thereby exposed this one; a fresh clone on arm64 always had it.

## Fix
The patched compiler maps `native`/`viper` to bytecode when `MICROPY_EMIT_NATIVE` is 0, so the optimization hint degrades gracefully. Ports **with** a native emitter (esp32 xtensa, unix x86_64) compile the fallback branch out entirely, so nothing changes there.

## Testing
- Rebuilt macOS target with the patch: `draw.py`, `blurhash.py`, and `nostr/nip44.py` all compile in the built binary (all three failed before).
- Full desktop boot: zero decorator errors, `NostrBootService` loads, zero tracebacks.
- `bash -n scripts/build_mpos.sh` passes; the guard skips cleanly when the patch file is absent (verified against the current pinned submodule SHA, which predates it).
- Not covered by a unit test: on CI x86_64 the native emitter exists, so the failing branch cannot be exercised there.

## Merge checklist
- CHANGELOG: ✅ entry under Future release → OS
- App version / MANIFEST: n/a
- MAINTAINERS.md / boards: n/a
- Settings migration: n/a
- Unit tests: not feasible on x86 CI (see above); reproduction + verification steps documented instead
- Docs: n/a (build-internal)
- After #6 merges, the lvgl_micropython submodule pointer bump will activate the patch; happy to open that bump PR too.


ThomasFarstrike:
> After https://github.com/MicroPythonOS/MicroPythonOS/issues/6 merges, the lvgl_micropython submodule pointer bump will activate the patch; happy to open that bump PR too.

I guess this was meant to reference a pull request (which has now been merged) instead of issue #6 ?

PR #240, opened 2026-08-09
main.py: skip stale lib/ override that bricks devices after OTA update

### LINKED ISSUE #239: Stale lib/ on flash shadows frozen modules after OTA update, crashing the launcher at boot
## Symptom

After an OTA update, the device shows the crash-warning dialog at every boot and never reaches a usable launcher:

```
App "com.micropythonos.launcher" threw an exception. It might be glitchy.
  File "lib/mpos/ui/view.py", line 56, in setContentView
  File "", line 1, in onResume
AttributeError: 'App' object has no attribute 'categories'
```

To a non-technical user this is a soft-brick: no launcher, no way to recover from the device itself, and the obvious remedy (re-flash via the web installer) wipes their prefs and installed apps.

## Root cause

- The web-installer LittleFS image is built with `mklittlefs -c internal_filesystem/` (scripts/mklittlefs.sh), so **every flashed device carries a full copy of `lib/` on the data partition** at whatever version the image was built.
- `main.py` unconditionally does `sys.path.insert(0, "lib")`, so that copy **shadows the frozen modules** — by design, as a development override.
- OTA updates replace the firmware partition (frozen framework + builtin apps) but never touch the data partition, so `lib/` stays at the originally-flashed version forever.
- The first OTA that crosses a framework API change makes the two halves incompatible. Concretely observed: firmware updated to **0.16.2** (builtin launcher reads `app.categories`, added in 0.16.1) while flash `lib/mpos` was still **0.16.0** (whose `App.__init__` doesn't set `categories`) → launcher crashes in `onResume` at boot.

A side effect that complicates support: `BuildInfo` is imported from the stale `lib/`, so the device *reports* the old version (0.16.0 here) while actually running 0.16.2 firmware — and the updater keeps offering the update it already applied.

## Diagnosis details (from the affected device)

- `BuildInfo.version.release` via normal import: `0.16.0`
- Same import with `lib` removed from `sys.path` and `mpos.*` purged from `sys.modules`: `0.16.2`
- `/builtin/apps/com.micropythonos.launcher/launcher.mpy` contains the string `categories`; `lib/mpos/app/app.py` on flash does not.
- No `apps/com.micropythonos.launcher` override was installed.

## Workaround (verified)

```
mpremote connect <port> resume exec "import os; os.rename('/lib', '/lib_stale')"
mpremote connect <port> resume exec "import machine; machine.reset()"
```

Device then boots the self-consistent frozen tree and reports the true firmware version. (`resume` matters: plain `exec` soft-resets on connect, which on some boards trips display re-init, MicroPythonOS#143.)

## Suggested fixes (complementary)

1. **Heal the installed base** — in the frozen `main.py`, before inserting `lib/` into `sys.path`, compare the release string in `lib/mpos/build_info.py` (plain text read, no import) against the frozen `BuildInfo`; on mismatch, skip `lib/` and log a prominent warning. Because `main.py` is frozen, this fix rides the OTA itself: every already-flashed device gets healed by its next update instead of bricked. PR incoming.
2. **Stop seeding the problem** — exclude `lib/` from the web-installer LittleFS image so end-user devices don't carry a shadowing copy at all; `lib/` remains a development override that developers create deliberately. (Left as a separate decision since it changes the image build and dev workflows.)

Happy to test patches on the affected device (Waveshare ESP32-S3 Touch LCD 2).

bitcoin3us:
Fixes #239.

## Summary

The web-installer LittleFS image ships a full copy of `lib/` on the data partition, and OTA updates only replace the firmware partition. `main.py` inserted `lib/` at the front of `sys.path` unconditionally, so the first OTA that crosses a framework API change leaves the stale `lib/mpos` shadowing the new frozen modules — and the OS crashes at boot. Observed in the wild as the 0.16.1+ launcher reading `app.categories` against a 0.16.0 `App` class:

```
AttributeError: 'App' object has no attribute 'categories'
```

To a non-technical user this is a soft-brick: crash dialog on every boot, no launcher, and re-flashing wipes their data. The stale `lib/` also makes `BuildInfo` report the old version, so the updater keeps offering the update that was already installed.

## Fix

Guard the `sys.path` insert in the frozen `main.py`:

- Read the release string out of `lib/mpos/build_info.py` **as plain text** (no import — stale code stays off the path).
- Compare with the frozen `BuildInfo`; the frozen import is purged from `sys.modules` afterwards so a *matching* `lib/` override still shadows `mpos` normally (the development-override feature is preserved).
- On mismatch, skip the `lib/` override and print a prominent warning; the device boots the self-consistent frozen tree.
- No `lib/mpos` on flash (no override installed), or no frozen `mpos` (bare interpreter): behavior unchanged.

Because `main.py` is frozen into the firmware, the fix is delivered **by the OTA update itself** — already-flashed devices get healed by their next update instead of bricked by it.

## Test plan

- [x] Guard exercised from the shipped source against the desktop binary in four scenarios: matching `lib/` → override kept; older `lib/` (0.16.0 and 0.9.9 fixtures) → skipped with warning; no `lib/` → kept. `sys.modules` verified free of cached `mpos.*` after every call.
- [x] Full boot through the modified `main.py` on the desktop binary (guard passes, `mpos.main` starts normally, no warning when versions match).
- [x] The underlying failure and the recovery (removing the stale `lib/`) reproduced and verified on a real bricked device (Waveshare ESP32-S3 Touch LCD 2, firmware 0.16.2 with `lib/` at 0.16.0).
- [ ] On-device test of the guard itself requires a firmware build containing it; happy to test a CI/esp32 build on the affected hardware.

## Note

This heals the installed base. The complementary longer-term change — not shipping `lib/` in the web-installer image at all — is left as a separate decision since it touches the image build and development workflows (see #239).


ThomasFarstrike:
That's an issue which I also bumped into, I think. And in your case the failure was pretty loud but it can also fail quietly, cause subtle bugs, causing you to wonder why...

PR #242, opened 2026-08-09
floodit: Tweaks, display "game ended" dialog and allow new game

pavelmachek:
This should make it less confusing. You can play more than one game now!


ThomasFarstrike:
Awesome!

ThomasFarstrike:
@pavelmachek would you mind also incrementing the version number? That way users will get an update notification :-)

pavelmachek:
I attempted to increase the version. Hopefully I got it right.

pavelmachek:
So... it worked, then I updated the version, and then it fails in unrelated test. I believe this is not my fault; tell me if you believe otherwise.

ThomasFarstrike:
Awesome, thanks! Indeed, that must be something else - hundreds of tests nowadays, and concurrent pull requests, need to structure this better. Will merge, thanks again, keep it coming!

PR #244, opened 2026-08-10
Status bar clock: stop allocating kilobytes per second

fdb:
The top-bar clock is the single largest steady allocator in the OS:

- update_time called mpos.time.localtime() THREE times per firing (once each for hours, minutes, seconds), and the timer fires every second.
- Every localtime() call ran the full POSIX-TZ pipeline in localPTZtime: string splits, filter lists, two transition parses

Together that is ~4,5 KB of allocations per second on an ESP32, causing excessive garbage collection pauses.

Three changes:

- topmenu.update_time calls localtime() once and indexes the tuple.
- localPTZtime grows a public tzoffset(timestamp, ptz) that returns (utcoffset, isdst, valid_from, valid_until); that allows us to cache TZ calculations properly, even during DST transitions.
- mpos.time.localtime() caches (ptz, offset, isdst, valid_from, valid_until) and reuses it while valid_from <= now < valid_until.

The cached path is one time.gmtime() plus one tuple, ~224 B. The full computation now runs about three times a year (two DST transitions and new year), plus on boot and on timezone change. We've done some extra work to make sure the validity interval is exact: the clock is correct to the second through DST transitions, and a backward clock jump (NTP correction) simply misses the cache and recomputes.

Verified against uncached tztime(): 409,690 comparisons across five zones (northern/southern DST, GMT0, angle-bracket half-hour offset, Julian-day rule) with second-by-second sweeps around every transition and year boundary 2025-2027, plus 1.64M randomized interval cross-checks 2020-2035: zero mismatches.

### DIFF
--- a/internal_filesystem/lib/localPTZtime.py
+++ b/internal_filesystem/lib/localPTZtime.py
@@ -115,27 +115,19 @@ def tziso(timestamp: float, ptz_string: str, zone_designator = True):
 	return stx
 
 
-def _timecalc(timestamp: float, ptz_string: str):
+def tzoffset(timestamp: float, ptz_string: str):
 	"""
-	Converts a time expressed in seconds in 10-tuple according to the time zone provided in Posix Time Zone format
+	Return the UTC offset in effect at 'timestamp' for the time zone provided in Posix Time Zone format, together with the interval of timestamps for which that offset is valid.
+	The offset only changes at a DST transition or when the year changes (transition rules are recomputed per year), so the result can be cached and reused for any timestamp within [valid_from, valid_until).
 
 	Parameters:
 	timestamp (float): Time in second
 	ptz_string (str): Time zone in Posix format
-	zone_designator (bool): Insert zone designator in returned string - default: True
-	
+
 	Returns:
-	tuple: 
-		* ``year``
-		* ``month``
-		* ``mday``
-		* ``hour``
-		* ``minute``
-		* ``second``
-		* ``weekday``
-		* ``yearday``
-		* ``isdst``
-		* ``utcoffset``
+	tuple: (utcoffset, isdst, valid_from, valid_until) where utcoffset is the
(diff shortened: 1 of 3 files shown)


ThomasFarstrike:
Oh waw, great find! These types of improvements are very much needed and appreciated!

ThomasFarstrike:
@fdb FYI, @jnuyens flagged a few more of these which I didn't get to yet:

- `ui/view.py:41,89` — focus group never `delete()`d (leak per nav); screen leaks on `remove_and_stop_all_activities` path (`clean()` != `delete()`; normal back path frees via `screen_load_anim(...,auto_del=True)` at `:109`). Small cumulative leak. Delete the popped group; delete screen on the remove-all path.
- `ui/font_manager.py:259,296` — TTF/emoji font caches never freed (no destroy calls) → C-heap fragmentation per app. Add `clear_cache()` on teardown.
- `ui/topmenu.py:299,423,502` — `create_notification_bar/drawer` have no re-entry guard; timers + focusable lists leak on re-create.
- `ui/file_explorer_activity.py:293` — action-bar focusables not removed from group on dismiss → deref of freed widgets.
- `task_manager.py:45` — `task_list` grows unbounded → RAM leak over session. Prune `done()` tasks. Note: it does seem good to leave them there for a while after exiting, just so a future "process manager" can see tasks ending, maybe.
- `app/activity.py:72` — `finish()` runs result callback after screen destroyed → use-after-free if it touches widgets. Note: not sure this is an issue, we're pretty resilient against these types of reference errors nowadays.
- `content/app_manager.py:700` — `get_services_for_action` leaves module in `sys.modules` (shadows later apps).

fdb:
Made PRs for the first two:

- focus group: https://github.com/MicroPythonOS/MicroPythonOS/pull/247
- font_manager: https://github.com/MicroPythonOS/MicroPythonOS/pull/248

Will see if I can do the other ones. Maybe better to put these as GH issues?

PR #247, opened 2026-08-11
Fix two LVGL memory leaks in activity navigation

fdb:
Navigation leaked memory on two paths:

1. Focus groups. Each `setContentView()` creates a focus group. LVGL keeps all groups in a global list, and the list is a GC root. The GC can never free a group; only `delete()` can. No code called it. Each navigation leaked ~90 bytes.
2. Screens on the remove-all path. `remove_and_stop_all_activities()` only called `clean()`. `clean()` deletes the children of a screen, but keeps the screen. Only the back path freed screens, through `screen_load_anim(..., auto_del=True)`. Thus `restart_launcher()` leaked the full screen stack on each app install or uninstall.

## Fix

- Delete the focus group when the activity is removed.
- Delete the screen when the activity is removed, with one exception. A screen that LVGL still tracks (active, previous, or loading) must not be deleted directly: delete only clears the active-screen pointer, so the other two would dangle. Keep that screen as `_orphan_screen`. The next screen load frees it with `auto_del`.

I've also added a regression test: It measures heap growth across ten navigations, and checks that both screens die on the remove-all path. Both tests fail on `main` and pass with this change.

## Known limit

A `remove_and_stop_all_activities()` call in the middle of a screen animation can still leak one screen. This falls back to the old behavior; it cannot crash.

### DIFF
--- a/internal_filesystem/lib/mpos/ui/view.py
+++ b/internal_filesystem/lib/mpos/ui/view.py
@@ -8,6 +8,11 @@ logger = logging.getLogger(__name__)
 
 screen_stack = []
 
+# Screen that outlived its activity because it was still on the display when
+# the activity was removed. It cannot be deleted right away, so the next
+# screen load frees it through LVGL's auto_del.
+_orphan_screen = None
+
 
 def close_top_layer_msgboxes():
     top = lv.layer_top()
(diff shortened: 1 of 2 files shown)


ThomasFarstrike:
Woop, woop!

PR #248, opened 2026-08-11
FontManager: free TTF and emoji fonts instead of leaking them per app

fdb:
`lv_tiny_ttf_create_file()` and `lv_imgfont_create()` allocate on the C heap,
which the MicroPython GC does not manage. `FontManager` cached both and never
called the matching destroy, so every app that loaded a TTF left it allocated
for the rest of the session and fragmented the C heap.

## Fix

Add `FontManager.clear_cache()`. It runs when an activity finishes and when all
activities stop.

The tricky part is figuring out which fonts are in use by a certain app, since destroying a font in use is a use-after-free bug which could crash the app. We do this by walking the layers, the screens a load animation runs between, and the activity screens on the stack,
and reads the text font of each object. We collect the pointers of these. A TTF is freed only when neither it nor
an emoji font composed on top of it is drawn, so a composed font is never left
with a dangling `fallback`.

Since this is a public API method, you should be able to call `clear_cache()` at any time withtout crashing the app. You can do this: it never frees memory LVGL still reads. Without that pointer check the method was a footgun: a label using a TTF, `clear_cache()`, then a redraw hung the test process outright. `clear_cache()` also returns early when no TTF is cached. If an app never loads a font, the method is free.

Builtin fonts stay cached. The builtin font list bounds how
many can exist, so they are not a per-app leak, and the persistent system UI on
`lv.layer_top()` holds live references to them.

## Matching fonts in use

To match a font in use, we're using the C pointer. The binding returns a fresh wrapper object each time, so `id()` never matches, and the wrapper's `__eq__` compares Python identity rather than the pointer:

    f.fallback = lv.font_montserrat_14
    f.fallback == lv.font_montserrat_14   # False

Instead, we use `bytes(font)`, which yields the raw `lv_font_t *`, which is the pointer to the font. We checked this by creating two TTFs from the same file and size — if it were the shared `get_glyph_dsc` field they would be identical:

    a bytes: 8081016c06000000
    b bytes: 2081016c06000000

## Tests 

`tests/test_graphical_font_manager.py` gains a `TestFontManagerClearCache` class
(8 tests), which tests clearing the cache after a font is no longer in use, and also tests that calling `clear_cache` randomly doesn't crash the app.


ThomasFarstrike:
This is great! It's these "less sexy" but SUPER important tasks that I never had time to look into yet!

I'll merge after the (final) [0.17.0 release](https://github.com/MicroPythonOS/MicroPythonOS/releases) is out :-)

ThomasFarstrike:
@fdb after merging this, the Meshcore app crashes when going into a chat channel, going back, and going into the chat channel again. I guess it's free'ing something that gets reused after, somehow...

@lucid-void FYI

fdb:
Hmm I’ll take a look tomorrow…On 16 Aug 2026, at 09:04, Thomas Farstrike ***@***.***> wrote:﻿ThomasFarstrike left a comment (MicroPythonOS/MicroPythonOS#248)
@fdb after merging this, the Meshcore app crashes when going into a chat channel, going back, and going into the chat channel again. I guess it's free'ing something that gets reused after, somehow...
@lucid-void FYI

PR #249, opened 2026-08-13
QR-code app deep links: scan a store link to install an app

bitcoin3us:
## What

Scan a QR code containing `https://apps.micropythonos.com/app/<app_id>` and land on that app's App Store detail screen, ready to install. Two entry points:

- **App Store**: new "Scan QR" top-bar button (reuses the CameraActivity `scanqr_intent` flow that WiFi settings already uses)
- **Camera app**: after decoding any QR in free-scan mode, an "Open in App Store" / "Open link" chip appears when the code is an app link the OS can open

The same URL doubles as a web landing page (companion PR in the `apps` repo), so phones scanning the QR get a useful page instead of an error — this is why the primary format is an HTTPS link rather than a custom scheme. `micropythonos://app/<id>` is accepted as an alias but never emitted.

## Security model

- **Identity, not location**: the QR carries only the app id; the `.mpk` is always downloaded from the trusted catalog, never from a URL in the QR
- Strict parser: exact host allowlist (no substring tricks), `https` only, userinfo/port rejected, charset + length caps, everything folded to lowercase (QR alphanumeric mode uppercases payloads)
- Scanning never auto-installs — the user confirms on the existing detail screen
- Third-party apps can declare URL handlers via `"urlPattern"` in manifest `intent_filters`; patterns matching the store host or the `micropythonos://` scheme are rejected at registration, so installed apps cannot shadow system links; multiple matching handlers open the existing ChooserActivity (no sticky default)

## Implementation

- `mpos/content/deeplink.py` (new): parser + `open_url()` dispatch, no LVGL/hardware imports so it stays desktop-testable
- `AppManager`: `_url_handler_specs` registry fed from manifests, `resolve_url_handlers()`
- AppStore `1.4.0`: scan button, deep link via `deeplink_fullname` intent extra, catalog refresh-on-miss, distinct dialogs for not-a-link / not-found / no-connection
- CameraActivity: action chip wiring (`deeplink.open_action_label()` decides the label)

## Testing

- `tests/test_deeplink.py`: 34 cases — hostile hosts, userinfo/port tricks, traversal, length caps, reserved-pattern registration, dispatch
- `tests/test_graphical_appstore_scanqr.py`: 6 cases (async work driven via captured TaskManager coroutines, per the `test_appstore_async_refresh.py` pattern)
- `tests/test_graphical_camera_qr_chip.py`: 6 cases (headless via `onCreate`, which builds the UI without camera hardware)
- All pass on the macOS desktop build alongside the existing appstore/intent/open-with suites (116 tests). Not yet verified on hardware — camera-through-the-lens testing still pending.

## Open question

The store index has been migrating to BadgeHub (the root index.html redirects there). Before this link format gets printed on anything physical, it's worth deciding whether QR links should stay on `apps.micropythonos.com`, whether BadgeHub URLs should join the device's host allowlist, or both.

Design doc with the full threat model and a rollout plan (`.mpk` signing gates any future third-party-repo QRs) available on request.


ThomasFarstrike:
Awesome!

Regarding:

> The store index has been migrating to BadgeHub (the root index.html redirects there).

Indeed, BadgeHub is becoming the default and apps.MicroPythonOS.com is being phased out.

Note that other backends, like https://upystore.io/ by @leezisheng might be getting added soon. They will have their own AppStore app, so we'll have "AppStore BadgeHub" and "AppStore uPyStore" or something like that, both visible in the main Launcher screen.

Does that change/help your architecture?

internal_filesystem/lib/mpos/content/app_manager.py
```diff
@@ -84,6 +88,40 @@ def _register_file_handler_spec(cls, action, app_fullname, entrypoint, classname
             "path_pattern": path_pattern,
         })
 
+    @classmethod
+    def _register_url_handler_spec(cls, app_fullname, entrypoint, classname, url_pattern):
+        """Store a manifest-declared URL handler after validating its pattern.
+
+        Patterns matching the official store host or the micropythonos://
```
ThomasFarstrike: Would it be possible to also add mpos:// where you use micropythonos:// ? Because that's about as long as http while micropythonos:// might be a bit too much typing work for some :-)

internal_filesystem/lib/mpos/content/deeplink.py
```diff
@@ -0,0 +1,272 @@
+"""App-link (deep link) parsing and dispatch.
+
+This module implements the QR-code app discovery link format:
+
+    https://apps.micropythonos.com/app/<app_id>[?v=<min_version>&s=<source>]
+    micropythonos://app/<app_id>[?...]           (accepted alias, never emitted)
```
ThomasFarstrike, line 6: I assume these are case insensitive, so MicroPythonOS:// would work as well? Is that being unit tested just to be sure? Thanks!

internal_filesystem/lib/mpos/content/app_manager.py
bitcoin3us: Done in f2841b0 — `mpos://app/<app_id>` now works everywhere `micropythonos://app/<app_id>` does, and both schemes are reserved against third-party `urlPattern` claims so neither can be shadowed by an installed app.

internal_filesystem/lib/mpos/content/deeplink.py
bitcoin3us: Yes — QR alphanumeric mode uppercases the whole payload, so the parser folds every part of the link (scheme, host, path prefix, app id) to lowercase. `MicroPythonOS://`, `MPOS://APP/COM.EXAMPLE.PAINT` etc. all parse. Now pinned by unit tests: `test_mixed_case_schemes` in tests/test_deeplink.py (added in f2841b0).

bitcoin3us:
> Note that other backends, like upystore.io might be getting added soon. They will have their own AppStore app ... Does that change/help your architecture?

It fits neatly, and mostly validates the split between the two mechanisms in this PR:

- **Per-store QR links** are exactly what the `urlPattern` manifest mechanism is for: "AppStore BadgeHub" declares `"urlPattern": "https://badgehub.eu/..."` in its intent_filters, "AppStore uPyStore" declares `upystore.io` patterns, and a scanned QR dispatches to the right store app automatically — no core changes needed per store. If two stores ever claim overlapping patterns, the existing ChooserActivity handles it.
- **The reserved namespace** stays minimal on purpose: only `apps.micropythonos.com` and the `micropythonos://`/`mpos://` schemes. BadgeHub/uPyStore hosts are intentionally *not* reserved, so store apps can own their links without touching the OS.
- **`mpos://app/<id>` is backend-neutral** — it names an app, not a store. Today it routes to the builtin AppStore; once multiple store apps exist, it could route to the user's default store or the chooser instead. That's a one-line change in `deeplink.open_url()`, so nothing in the link format needs to change when the backend landscape shifts.

As apps.micropythonos.com gets phased out, the printed-QR question becomes: what URL should a *store-agnostic* app QR carry? `mpos://app/<id>` works on-device forever but does nothing on phones; an HTTPS link needs a host that will keep existing. Happy to adjust the allowlist/dispatch whichever way you land on that.

ThomasFarstrike:
Merging!

PR #251, opened 2026-08-15
Fix UNIHIKER K10 two-button navigation

leezisheng:
- add K10-specific short-press select and long-press back handling
- use B to move widget or keyboard focus with visible key highlighting
- keep Launcher and HowTo controls reachable on the two-button device
- cover the K10 button state machine with unit tests

### DIFF
--- a/internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
+++ b/internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
@@ -49,7 +49,7 @@ class HowTo(Activity):
         if DeviceInfo.get_hardware_id() == "unihiker_k10":
             # K10 uses B as its only NEXT key, so static help text stays out of
             # its focus cycle and must remain compact enough to fit on one screen.
-            self._add_label(screen, "B: next. Hold B in a list: previous.", focusable=False)
+            self._add_label(screen, "Open a drop-down: B next. Hold B: previous.", focusable=False)
             self._add_label(screen, "A: select. Hold A: cancel or go back.", focusable=False)
         elif InputManager.has_pointer():
             self._add_label(screen, touchhelp)
--- a/internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
+++ b/internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
@@ -1,6 +1,6 @@
 import logging
 
-from mpos import Activity, SharedPreferences, DisplayMetrics, add_focus_highlight
+from mpos import Activity, DeviceInfo, SharedPreferences, DisplayMetrics, add_focus_highlight
 import lvgl as lv
 
 logger = logging.getLogger(__name__)
@@ -10,6 +10,7 @@ class HowTo(Activity):
     appname = "com.micropythonos.howto"
 
     dontshow_checkbox = None
+    closebutton = None
     prefs = None
 
     def onCreate(self):
@@ -17,22 +18,28 @@ class HowTo(Activity):
         screen.set_style_border_width(0, lv.PART.MAIN)
         screen.set_flex_flow(lv.FLEX_FLOW.COLUMN)
         screen.set_style_pad_all(DisplayMetrics.pct_of_width(5), lv.PART.MAIN)
-        # Children (labels, checkbox, button) are auto-added to the default group;
-        # the screen itself must NOT be in the group, otherwise move_focus_direction
-        # cannot find widgets below it (the fullscreen rect blocks spatial navigation).
-        preamble = "How to Navigate"
+        # Only actionable controls should go into the focus group.
+        preamble = "Navigate"
         self._add_label(screen, preamble, is_header=True)
 
-        buttonhelp_intro = "As you don't have a touch screen, you need to use the buttons to navigate:"
         buttonhelp_items = [
-            "If you have a joystick and at least 2 buttons, then use the joystick to move around. Use one of the buttons to ENTER and another to go BACK.",
-            "If you have 3 buttons, then one is PREVIOUS, one is ENTER and one is NEXT. To go back, press PREVIOUS and NEXT together.",
-            "If you have just 2 buttons, then one is PREVIOUS, the other is NEXT. To ENTER, press both at the same time. To go back, long-press the PREVIOUS button.",
+            "Joystick: move. Btns: ENTER/BACK.",
+            "3 buttons: PREV/ENTER/NEXT.",
+            "2 buttons: PREV/NEXT.",
         ]
-        touchhelp = "Swipe from the left edge to go back and from the top edge to open the menu."
+        buttonhelp_intro = "Use buttons to navigate:"
+        touchhelp = "Left/back. Menu top."
         from mpos import InputManager
+        try:
+            is_k10 = DeviceInfo.get_hardware_id() == "unihiker_k10"
+        except Exception:
+            is_k10 = False
         if InputManager.has_pointer():
             self._add_label(screen, touchhelp)
+        elif is_k10:
+            self._add_label(screen, "K")
+            self._add_label(screen, "- B: next.")
+            self._add_label(screen, "- A: select.")
         else:
             self._add_label(screen, buttonhelp_intro)
             for item in buttonhelp_items:
@@ -40,12 +47,15 @@ class HowTo(Activity):
 
         self.dontshow_checkbox = lv.checkbox(screen)
         self.dontshow_checkbox.set_text("Don't show again")
+        add_focus_highlight(self.dontshow_checkbox)
 
         closebutton = lv.button(screen)
+        self.closebutton = closebutton
         closebutton.add_event_cb(lambda *args: self.finish(), lv.EVENT.CLICKED, None)
         closebutton.add_event_cb(self._on_long_press, lv.EVENT.LONG_PRESSED, None)
         closelabel = lv.label(closebutton)
         closelabel.set_text("Close")
+        add_focus_highlight(closebutton)
 
         self.setContentView(screen)
 
@@ -54,8 +64,6 @@ class HowTo(Activity):
         label.set_width(lv.pct(100))
         label.set_text(text)
         label.set_long_mode(lv.label.LONG_MODE.WRAP)
-        add_focus_highlight(label)
-        lv.group_get_default().add_obj(label)
         if is_header:
             label.set_style_text_font(lv.font_montserrat_24, lv.PART.MAIN)
             label.set_style_margin_bottom(4, lv.PART.MAIN)
@@ -76,6 +84,12 @@ class HowTo(Activity):
             self.dontshow_checkbox.remove_state(lv.STATE.CHECKED)
         else:
             self.dontshow_checkbox.add_state(lv.STATE.CHECKED)
+
+        if self.closebutton:
+            try:
+                lv.group_focus_obj(self.closebutton)
+            except Exception:
+                pass
 
     def onPause(self, screen):
         if __debug__: logger.debug("howto app onPause called")
--- a/internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
+++ b/internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
@@ -18,33 +18,47 @@ class HowTo(Activity):
         screen.set_style_border_width(0, lv.PART.MAIN)
         screen.set_flex_flow(lv.FLEX_FLOW.COLUMN)
         screen.set_style_pad_all(DisplayMetrics.pct_of_width(5), lv.PART.MAIN)
-        # Only actionable controls should go into the focus group.
-        preamble = "Navigate"
-        self._add_label(screen, preamble, is_header=True)
+        self._add_label(
+            screen,
+            "How to Navigate",
+            is_header=True,
+            focusable=DeviceInfo.get_hardware_id() != "unihiker_k10",
+        )
 
         buttonhelp_items = [
-            "Joystick: move. Btns: ENTER/BACK.",
-            "3 buttons: PREV/ENTER/NEXT.",
-            "2 buttons: PREV/NEXT.",
+            (
+                "If you have a joystick and at least 2 buttons, then use the joystick "
+                "to move around. Use one of the buttons to ENTER and another to go BACK."
+            ),
+            (
+                "If you have 3 buttons, then one is PREVIOUS, one is ENTER and one "
+                "is NEXT. To go back, press PREVIOUS and NEXT together."
+            ),
+            (
+                "If you have just 2 buttons, then one is PREVIOUS, the other is NEXT. "
+                "To ENTER, press both at the same time. To go back, long-press the "
+                "PREVIOUS button."
+            ),
         ]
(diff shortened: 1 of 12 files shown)


leezisheng:
Fix UNIHIKER K10 navigation, camera orientation, and input recovery
## Summary

  Improve MicroPythonOS support for the DFRobot UNIHIKER K10.

  - add K10-specific two-button navigation
  - use a short A press to activate the focused control
  - use a long A press to return to the previous screen
  - use B to cycle through focusable controls and keyboard entries
  - keep Launcher, HowTo, and Camera controls reachable without a touchscreen
  - rotate the K10 display by 180 degrees to match the physical board orientation
  - correct the GC2145 RGB565 byte order and default camera orientation
  - restore the XL9535 I2C bus after camera initialization and settings changes
  - preserve A/B button input after the camera takes control of the shared GPIO47/48 bus
  - add per-camera RGB565 byte-order and vertical-flip metadata

  ## Scope

  The display rotation and XL9535 bus recovery are implemented only in the
  UNIHIKER K10 board configuration.

  Other cameras retain the existing defaults:

  - standard RGB565 byte order
  - vertical flip enabled by default

  Camera controls are now added to the LVGL focus group so boards using keypad
  or encoder input can operate the Camera app.

  ## Validation

  - K10 input and XL9535 recovery tests: passed
  - CameraManager tests: passed
  - graphical camera format and focus tests: passed
  - syntax tests: passed
  - `make lint`: passed
  - ESP32-S3 full firmware build: passed
  - full firmware image flashed successfully to a UNIHIKER K10

  ## Hardware validation


a72f638b650959e06efaf34abc6e783e

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
```diff
@@ -10,42 +10,52 @@ class HowTo(Activity):
...
-            "If you have a joystick and at least 2 buttons, then use the joystick to move around. Use one of the buttons to ENTER and another to go BACK.",
-            "If you have 3 buttons, then one is PREVIOUS, one is ENTER and one is NEXT. To go back, press PREVIOUS and NEXT together.",
-            "If you have just 2 buttons, then one is PREVIOUS, the other is NEXT. To ENTER, press both at the same time. To go back, long-press the PREVIOUS button.",
+            "Joystick: move. Btns: ENTER/BACK.",
+            "3 buttons: PREV/ENTER/NEXT.",
+            "2 buttons: PREV/NEXT.",
         ]
-        touchhelp = "Swipe from the left edge to go back and from the top edge to open the menu."
+        buttonhelp_intro = "Use buttons to navigate:"
+        touchhelp = "Left/back. Menu top."
```
ThomasFarstrike: I would propose the punctuation: Left: back. Top: menu.

But I'm not sure that beginning users would understand "left" to mean "swipe from the left edge".
Same for "top" - they might think they have to tap at the top, without realizing they have to "swipe from the top"... For us that's very obvious but for new users, that might not be expected...

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
```diff
@@ -10,42 +10,52 @@ class HowTo(Activity):
...
         self.dontshow_checkbox.set_text("Don't show again")
+        add_focus_highlight(self.dontshow_checkbox)
 
         closebutton = lv.button(screen)
+        self.closebutton = closebutton
         closebutton.add_event_cb(lambda *args: self.finish(), lv.EVENT.CLICKED, None)
         closebutton.add_event_cb(self._on_long_press, lv.EVENT.LONG_PRESSED, None)
         closelabel = lv.label(closebutton)
         closelabel.set_text("Close")
+        add_focus_highlight(closebutton)
```
ThomasFarstrike, line 72: Why is this needed? Buttons are added to the focus group and have the focus highlight (border) by default, I think...?

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
```diff
@@ -77,6 +85,12 @@ def onResume(self, screen):
         else:
             self.dontshow_checkbox.add_state(lv.STATE.CHECKED)
 
+        if self.closebutton:
+            try:
+                lv.group_focus_obj(self.closebutton)
```
ThomasFarstrike: Maybe add a comment here, like: # focus on the close button by default for convenience

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
```diff
@@ -10,42 +10,52 @@ class HowTo(Activity):
...
+        # Only actionable controls should go into the focus group.
+        preamble = "Navigate"
         self._add_label(screen, preamble, is_header=True)
 
-        buttonhelp_intro = "As you don't have a touch screen, you need to use the buttons to navigate:"
         buttonhelp_items = [
-            "If you have a joystick and at least 2 buttons, then use the joystick to move around. Use one of the buttons to ENTER and another to go BACK.",
-            "If you have 3 buttons, then one is PREVIOUS, one is ENTER and one is NEXT. To go back, press PREVIOUS and NEXT together.",
-            "If you have just 2 buttons, then one is PREVIOUS, the other is NEXT. To ENTER, press both at the same time. To go back, long-press the PREVIOUS button.",
+            "Joystick: move. Btns: ENTER/BACK.",
```
ThomasFarstrike: Is it necessary to use "Btns" instead of "Buttons" ? Seems like a small difference...

Also, the "move" action is lower case while the ENTER/BACK is all caps.

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
```diff
@@ -10,42 +10,52 @@ class HowTo(Activity):
...
+            "3 buttons: PREV/ENTER/NEXT.",
+            "2 buttons: PREV/NEXT.",
         ]
-        touchhelp = "Swipe from the left edge to go back and from the top edge to open the menu."
+        buttonhelp_intro = "Use buttons to navigate:"
+        touchhelp = "Left/back. Menu top."
         from mpos import InputManager
+        try:
+            is_k10 = DeviceInfo.get_hardware_id() == "unihiker_k10"
```
ThomasFarstrike: is it necessary to have is_k10 be a separate variable? Maybe it's slightly simpler to just use:

`elif DeviceInfo.get_hardware_id() == "unihiker_k10":`

directly below?

internal_filesystem/lib/mpos/ui/camera_activity.py
```diff
@@ -156,6 +157,23 @@ def resize_buttons(self):
...
+        for button in (
+            self.close_button,
+            self.settings_button,
+            self.qr_button,
+            self.snap_button,
+        ):
+            mpos_ui.add_focus_highlight(button, width=2)
+        lv.group_focus_obj(self.close_button)
+
+    def _get_camera_defaults(self, mode_defaults):
```
ThomasFarstrike, line 175: I think one line of comment above this function would be nice, especially since it was there before:

`# Merge common and scanqr-specific defaults`

internal_filesystem/lib/mpos/camera_manager.py
```diff
@@ -40,7 +40,7 @@ class Camera:
     Represents a camera device with its characteristics.
     """
 
-    def __init__(self, lens_facing, name=None, vendor=None, version=None, init=None, deinit=None, capture=None, apply_settings=None, rotation_degrees=0):
+    def __init__(self, lens_facing, name=None, vendor=None, version=None, init=None, deinit=None, capture=None, apply_settings=None, rotation_degrees=0, rgb565_byte_swap=False, default_vflip=True):
```
ThomasFarstrike, line 43: Nice one!

ThomasFarstrike (changes requested):
Thanks, looks good!

I added a few review comments, I hope that helps.

Also, could you please give some background info on why the text in the HowTo app was changed? I think it will be a bit harder to understand for beginning users...
I understand you wanted to make it shorter to fit the screen but the user can also scroll by focusing on the next text label by pressing "next", I think. Or not?
If you really need to have it so short, then at least we should have a comment that warns the next developer to keep the text short for reason X or Y. Otherwise they will make it longer again :-)

Anyway, it's very nice, thank you for taking the time to work on this!

EDIT: I forgot to mention; feel free to add a line to the "CHANGELOG.md" about this pull request so that users will know what's been changed/improved before deciding to update/install the next release :+1:

leezisheng:
Thanks for the careful review. You are right that I shortened the HowTo text too aggressively while keeping the K10 frozen firmware within its tight app partition budget. That made the guidance
  less clear for beginners and also affected devices beyond the K10.

  I will restore the beginner-friendly generic wording, make the compact two-button guidance K10-only, and explicitly explain B = next, short A = select, and long A = back. I will also clarify the
  swipe gestures, add the requested comments, and add a CHANGELOG entry.

  I also found an additional K10 issue while retesting: the AppStore category dropdown receives NEXT instead of RIGHT, so B moves focus away instead of operating the dropdown. I will add a K10-
  only dropdown mapping and regression coverage before updating this PR.

  The Unix test failures also occur on the PR base and current main, so I will not present CI as green. I will rebase onto current main, rerun the relevant checks, and report the exact remaining
  CI status.

leezisheng:
Yes, non-K10 devices can scroll by focusing text labels. On K10, however, B is the only NEXT key, so traversing every static label is cumbersome. I will restore the full generic text and keep
     static labels out of the focus group only for K10, with a complete compact K10 instruction set and a code comment explaining this constraint.

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
leezisheng: Agreed. I will restore explicit swipe wording rather than use the shorthand

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
leezisheng: It is needed here because add_focus_highlight() explicitly registers the actionable control in the default LVGL focus group and installs the MPOS focus-border handlers. I will add a short
     comment to make that intent clear.

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
leezisheng: Agreed. I will scope this to K10 and add a comment explaining that Close is focused first so a two-button user can exit immediately.

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
leezisheng: Agreed. I will use “Buttons” and consistent sentence-style wording. The compact wording will be K10-only.

internal_filesystem/builtin/apps/com.micropythonos.howto/howto.py
leezisheng: Agreed. DeviceInfo.get_hardware_id() is safe here, so I will use the direct condition and remove the intermediate variable.

internal_filesystem/lib/mpos/ui/camera_activity.py
leezisheng: Agreed. I will add a concise comment describing the common, mode-specific, and board-specific defaults merged by this helper.

internal_filesystem/lib/mpos/camera_manager.py
leezisheng: Thank you. The metadata keeps the K10 RGB565 and orientation correction opt-in, while other boards retain their existing defaults.

leezisheng:
This follow-up scopes the camera focus setup to UNIHIKER K10. The K10 has no touch screen, joystick, or directional control; it only has the A and B buttons. Its board input layer therefore
  needs a stable initial camera action and visible focus feedback. Applying that setup unconditionally in CameraActivity would change the initial focus behavior on other boards, so non-K10 boards
  now keep their existing focus behavior.

  LVGL already adds buttons to the default focus group. The K10-specific code only adds the explicit focus highlight and selects Close as the initial target. The regression test now covers both
  cases: K10 starts with Close focused, while a non-K10 board preserves its existing focused control.

  I also documented the camera metadata and K10-specific hardware behavior: RGB565 byte order, default vertical flip, display rotation, shared SCCB/XL9535 I2C recovery, and the Launcher choice to
  resume on a launchable app instead of HowTo. These comments are intended to prevent future camera or input changes from reintroducing the K10 regressions.

  Validation completed with the focused CameraManager, camera color-format, K10 input, AppStore, HowTo, and Launcher tests (54 tests total), syntax checks, targeted Ruff checks, and an ESP32-S3
  firmware build. The resulting app image remains within the smallest partition, with 10,992 bytes free.

leezisheng:
## Summary

  - restore button input after camera use and correct K10 camera preview orientation
  - make Launcher, HowTo, AppStore dropdowns, and Camera usable with only A and B buttons
  - keep CameraActivity's explicit focus setup specific to UNIHIKER K10
  - document K10 camera metadata, RGB565 byte order, display orientation, and shared I2C recovery
  - add regression coverage for K10 and non-K10 camera focus behavior

  ## Why

  UNIHIKER K10 has no touch screen, joystick, or directional control. It only has the A and B buttons, so camera controls need visible focus feedback and a predictable initial target.

  LVGL already adds buttons to its default focus group. The K10-specific code only adds the explicit focus highlight and selects Close as the initial target. Other boards retain their existing
  focus behavior.

  ## Validation

  - 54 focused CameraManager, camera color-format, K10 input, AppStore, HowTo, and Launcher tests passed
  - `make syntax-tests`, targeted Ruff, and `git diff --check` passed
  - ESP32-S3 image built successfully: 3,659,024 bytes, 10,992 bytes free in the smallest app partition
  - flashed the UNIHIKER K10 app partition; esptool hash verification and on-device verification completed

internal_filesystem/builtin/apps/com.micropythonos.launcher/launcher.py
```diff
@@ -202,6 +202,7 @@ def _exit_splash_mode(self, screen):
     def _focus_first_usable_app(self):
         for fullname, target in self._app_cont_map.items():
             if fullname == "com.micropythonos.howto":
+                # K10 should resume on a launchable app instead of the startup HowTo.
```
ThomasFarstrike: This comment mentions "K10" but it's applicable for all boards, not only the K10, right?

So maybe just use something like: `# don't focus on the HowTo app because <put reason here>`

ThomasFarstrike (changes requested):
Thank you for updating! I just had one little tiny comment, if you don't mind. If you disagree, please don't hesitate to say so. Looking forward to merging this!

internal_filesystem/builtin/apps/com.micropythonos.launcher/launcher.py
leezisheng: Thanks, good catch. The helper itself looks generic, but its only call sites are guarded by self._k10, so other boards keep their existing focus behavior. I will reword the comment to make that K10-only fallback explicit.

ThomasFarstrike:
Perfect, merged!

PR #258, opened 2026-08-17
FontManager: free an app's TTF and emoji fonts when the app quits

fdb:
Second attempt at #248, which was reverted in #254 because the MeshCore app crashed when re-entering a chat channel.

## The bug in #248

I looked at the #248 implementation and at how MeshCore uses fonts, and found the issue: MeshCore caches the font handle in a module global and applies it each time a chat channel screen is built. #248's `clear_cache()` freed fonts whenever an activity finished, scanning the widget tree to decide what was still in use. But a widget scan only sees fonts currently applied to a widget — it cannot see a handle an app keeps in a Python variable for later. So:
enter channel → back (font freed) → enter channel again (freed font applied to a new label) → use-after-free.

## The root cause

MeshCore's caching the font in a global is normal app code, not some weird edge case, so we absolutely need to support this. The real problem is that `getFont()` returns an object with no explicit lifetime contract, held forever by FontManager's own cache: `_ttf_font_cache` is a class attribute on an OS-lifetime singleton, so every font any app ever loaded stays reachable for the rest of the session. The app's own references die with the app — the OS's cache reference is what pins the memory.

A deep dive into lvgl_micropython showed that the fonts themselves are already managed: LVGL allocates from the MicroPython GC heap (`lv_conf.h` sets `LV_USE_STDLIB_MALLOC = LV_STDLIB_MPY`), and the conservative GC follows C-side references too — a style or widget that holds an `lv_font_t*` keeps that font alive. We verified this empirically: a font referenced only by a C-side style survives `gc.collect()`, and is reclaimed after `style.reset()`, with no destroy call.

## The fix

Since the GC already tracks who still holds a font, we only drop the cache references when the app quits. That happens in two places — `finish_current_activity()` when the stack drops to launcher-only, and `remove_and_stop_all_activities()`. The GC then reclaims each font once nothing uses it, and keeps every font that an app, a widget, or a thread still holds. Use-after-free is impossible by construction: we never free memory; the GC frees only what is unreachable. Fonts shared by stacked apps (same TTF path and size) are dropped only once all of them are gone.

This also made the code simpler:

- No destroy calls. `_clear_cache()` removes dictionary entries; `tiny_ttf_destroy()` / `imgfont_destroy()` are never called.
- The widget-tree walk is gone. It existed to decide when freeing was safe. Nothing is freed anymore, so there is nothing to decide.
- `_get_ttf_font()` reads the .ttf into memory and calls `tiny_ttf_create_data()` instead of `tiny_ttf_create_file()`. This is load-bearing: a file-based font keeps an `lv_fs` file open that only a destroy call closes — the one resource the GC cannot see. A data-based font keeps its buffer alive through the GC, so both are reclaimed together.
- The method is private (`_clear_cache()`): it has one caller, the OS teardown path in `view.py`.
- The `getFont()` docstring documents the contract: fonts are cached; the OS drops an app's custom fonts from the cache when the app quits; a font you still hold stays valid — from anywhere: activities, services, threads, LVGL callbacks, "Open With" handlers.

Again the goal of #248 still holds: avoid font leaks and heap fragmentation for apps that use custom fonts — by dropping the cache references so the GC reclaims the fonts once they're no longer in use.

## Tests

New regression test `TestFontManagerAppHeldFont` replays the MeshCore sequence: font held in a variable, sub-activity finished, font re-applied and rendered, then dropped from the cache once the app's last activity is gone. It fails against the #248 behavior (3/3 runs) and passes with this change. All 64 font manager tests and the activity-stack navigation tests pass.

On the desktop emulator with the real MeshCore app from BadgeHub, channel → back → channel survives (the flow that crashed #248 with SIGBUS), and quitting to the launcher lets the GC reclaim ~110 KB. `tests/manual_test_meshcore_font_reentry.py` replays this flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ Future release (next version)
 =====
 
 Frameworks:
-- FontManager: delete an app's TTF and emoji fonts when the app closes, instead of leaking them, by @fdb
+- FontManager: stop leaking an app's TTF and emoji fonts after the app closes, by @fdb
 
 0.17.3
 ======
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Future release (next version)
 =====
 
+Frameworks:
+- FontManager: free TTF and emoji fonts once the last app activity closes, instead of leaking them per app, by @fdb
+
 0.17.3
 ======
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ Future release (next version)
 =====
 
 Frameworks:
-- FontManager: free TTF and emoji fonts once the last app activity closes, instead of leaking them per app, by @fdb
+- FontManager: delete an app's TTF and emoji fonts when the app closes, instead of leaking them, by @fdb
 
 0.17.3
 ======
--- a/internal_filesystem/lib/mpos/ui/font_manager.py
+++ b/internal_filesystem/lib/mpos/ui/font_manager.py
@@ -102,15 +102,21 @@ lips,🫦
 
     @classmethod
     def getFont(cls, size=None, ttf=None, family=None, emoji=False):
-        """Return a font, loading and caching it on first use.
+        """Return a font for use in this app's widgets.
 
-        Lifetime: a font loaded with ttf= (and any emoji font composed on it)
-        is valid until the last activity of the running app closes. The OS
-        destroys it then, so the C heap it lives on is not leaked. Within an
-        app run you can keep and reuse the returned font freely, but do not
-        cache it in a service or thread that outlives your activities, and do
-        not apply it to OS-owned widgets such as lv.layer_top(). Builtin and
-        emoji-composed builtin fonts are never destroyed.
+        size picks the closest builtin size, family picks a builtin family
+        (e.g. "Montserrat"), ttf loads a custom font from a .ttf file, and
+        emoji=True adds emoji support on top of the chosen font.
+
+        Note: only use custom (ttf=) fonts in your activities. When your app
+        closes, the OS deletes its custom fonts, so the device does not
+        slowly run out of memory. Code that keeps running after your app
+        closes — a service, a thread, a timer, an LVGL callback — must not
+        hold on to a font: the font it remembers is gone, and drawing with
+        it crashes the device. For the same reason, do not put a custom font
+        on a system widget like lv.layer_top(), because it stays on screen
+        after your app closes. Fonts are cached, so calling getFont() again
+        is cheap. Builtin fonts are always safe.
         """
         target_size = cls._normalize_size(size)
 
--- a/internal_filesystem/lib/mpos/ui/font_manager.py
+++ b/internal_filesystem/lib/mpos/ui/font_manager.py
@@ -136,20 +136,21 @@ lips,🫦
 
     @classmethod
     def _clear_cache(cls):
-        """Destroy every cached TTF font and the emoji imgfonts composed on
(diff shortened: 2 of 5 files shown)


ThomasFarstrike:
Nice, thanks for giving it another go!

Just a follow-up with regards to:

> a finalizer in the lvgl_micropython bindings would be the fundamental fix — lifetime would follow Python reachability — but that is an upstream binding change, out of scope here.

Why is lvgl_micropython out of scope?

We already fork it https://github.com/MicroPythonOS/lvgl_micropython and carry a bunch of patches (unfortunately) in .patch files on the "integration" branch or in "topic/topicname" branches which are merged into integration...

Alternatively, or additionally, a pull request could be submitted to lvgl_micropython.

Maybe a first step would be to just attach a .patch file here that fixes the issue in lvgl_micropython itself, and then we can decide if it's clean/simple enough to take into consideration?

I appreciate the "Known limitations" section!

fdb:
Hey @ThomasFarstrike, thanks for the review and sorry for the mess with the previous PR! The MeshCore example showed that there were different ways fonts could be used, so I had to change my approach. 

You're right, adding a patch to lvgl_micropython to track lifetimes might be an option, that might make this safer as well. I'll see if that's possible as well.

I've left this PR as a draft since I'm still testing regressions, and trying out in real-world scenarios (e.g. with MeshCore) to see if the behavior is fixed with this PR.

fdb:
The suggestion to do a deep dive into lvgl_micropython was useful: it turned out the binding *already* manages font lifecycles, which made this implementation simpler.

So: LVGL allocates from the MicroPython GC heap: lv_conf.h sets LV_USE_STDLIB_MALLOC = LV_STDLIB_MPY, which routes lv_malloc through m_malloc, and the binding registers LVGL's roots with the GC. Conservative marking thus follows C-side references too: a style or widget that holds an lv_font_t* keeps that font alive.

**The root cause**: the leak was never unmanaged C memory; it was FontManager's own global cache: `_ttf_font_cache` is a class attribute on an OS-lifetime singleton, so every font loaded by any app stayed reachable forever. So, to avoid gobbling up memory, an OS-lifetime cache must not hold app-lifetime objects forever. Dropping the cache entries when the app quits is the solution.

## Solution

- `_clear_cache()` only drops the cache references. No more destroy calls: the GC reclaims each font once nothing uses it, and keeps every font that an app, a widget, a thread, or a style still holds. **Use-after-free becomes impossible** by construction: we never free memory, and the GC frees only what is unreachable.
- `_get_ttf_font()` now reads the .ttf file into memory and calls tiny_ttf_create_data() instead of tiny_ttf_create_file(). We do this because a file-based font keeps an lv_fs file open that only a destroy call closes, so that's a resource the GC cannot see. With data-based fonts, the font keeps its buffer alive and both are reclaimed together. The cost is the font file in RAM (tens of KB for a subset font); the glyph caches cost a similar amount anyway.
- Wording in the code: we don't talk about destroying fonts but dropping fonts; the GC decides when the memory returns.

**Good news**: The known-limitations list from the PR description is obsolete: a font held by a service, a thread, an LVGL callback, or an "Open With" handler simply stays alive until its holder is gone.

## Verification

All 64 font manager tests pass. The real MeshCore app on the desktop emulator survives channel → back → channel (the flow that crashed #248), and quitting to the launcher drops the cache and lets the GC reclaim ~110 KB — without a single destroy call (just `drop`). tests/manual_test_meshcore_font_reentry.py recreates this flow to avoid regressions.

ThomasFarstrike:
Oof, this sounds perfect. Thanks for digging deep into it! Let's merge it and see!

PR #260, opened 2026-08-19
Add CTRL-SHIFT-S screenshot hotkey to the desktop emulator

fdb:
This PR adds a hotkey to SDL to make a screenshot from the current app. This captures the actual pixels, not the scaled version. 

We re-used the BMP code that the web server already had and moved that into a shared function (`mpos.ui.testing.encode_bmp()`), instead of copying the BMP write code.

So CTRL-SHIFT-S (CMD-SHIFT-S on macOS) saves screenshot-YYYYMMDD-HHMMSS.bmp in the `internal_filesystem` directory and prints the absolute path.

This is for your consideration; if you want a different shortcut key, let me know and I can change it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ conf.json*
 # auto created when running on desktop:
 internal_filesystem/SDLPointer_2
 internal_filesystem/SDLPointer_3
+internal_filesystem/screenshot-*.bmp
 
 internal_filesystem.mpy/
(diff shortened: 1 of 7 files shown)


ThomasFarstrike:
Haha, that's cool! Thanks for restructuring the code and making sure the BMP logic is not duplicated in 2 places!

PR #264, opened 2026-08-21
Cache emoji codepoints for keyboard input

fdb:
This PR avoids rebuilding the emoji codepoint set on every regular keyboard input.

Previously, `MposKeyboard._contains_emoji()` called `FontManager.getEmojiCodepoints()` for every character until the user entered their first emoji. That method walked the full emoji map, split and parsed every key, built a set, and sorted it into a list every time you pressed a key.

We added `FontManager.isEmojiCodepoint()` for the keyboard hot path. This checks the cached set directly, so regular keyboard input now does one `ord()` and one set lookup without rebuilding any collections. `getEmojiCodepoints()` still returns a new sorted list, preserving its existing API. The emoji codepoint bounds calculation also reuses the cached set instead of parsing the map keys again.

I added tests that cover identifying available emoji, rejecting regular characters, and reusing the same cached codepoint index across calls.

### DIFF
--- a/internal_filesystem/lib/mpos/ui/font_manager.py
+++ b/internal_filesystem/lib/mpos/ui/font_manager.py
@@ -22,6 +22,8 @@ class FontManager:
     # probes, and repeated per-codepoint fallback walks.
     _emoji_map = None  # dict of hex-key -> src path, populated on first use
     _emoji_strings = None  # list of complete emoji strings, populated on first use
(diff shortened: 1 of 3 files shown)


ThomasFarstrike:
It's really great to have someone taking a critical look and finding what I'm sure must be horrible performance drains that are hiding in the code. Thank you!

PR #265, opened 2026-08-21
Release completed TaskManager tasks

### LINKED ISSUE #261: TaskManager retains completed tasks indefinitely (memory leak)
`TaskManager` keeps a strong reference to every task ever created, including tasks that have completed.

In `task_manager.py`, the `task_list` is a class-level variable, and `create_task()` only appends to it, with no cleanup path. (There is actually a comment that says so: `# might be good to periodically remove tasks that are done, to prevent this list from growing huge`.)

Tasks like update checks, UI actions, etc. can grow this list unbounded.

## Example

Real application tasks also reference the coroutine, so also its frame, local variables and owner object. So if we give every coroutine a 10KB buffer:

1. Create 100 short-lived coroutines through `TaskManager.create_task()`.
3. Yield to the event loop until every task completes.
4. Run garbage collection and inspect `task_list` and free heap memory.

Observed:

- `len(TaskManager.task_list) == 100`
- All 100 tasks report `done() == True`
- 1,034,944 bytes (approximately 1.03 MB) remain retained after garbage collection

If we prune completed tasks (see the proposed fix below), we no longer retain any tasks and thus release all memory.

## Severity

Initially I thought this would grow on the badge just being idle, but that's not the case. App store and OS udpate tasks have persistent coroutines, so they don't create new tasks. 

The issue is larger for apps that create tasks, not realizing that this costs memory even after the task is done. E.g. Nostr/ESPNow chat create approximately one task per sent message.

## Proposed fix

`create_task` could discard completed tasks before creating a new task, preserving tasks that are still running:

```python
    @classmethod
    def create_task(cls, coroutine):
        cls.task_list = [task for task in cls.task_list if not task.done()]
        task = asyncio.create_task(coroutine)
        cls.task_list.append(task)
        return task
```

fdb:
## Summary

Fixes #261.

`TaskManager.create_task()` now runs managed coroutines through a cleanup wrapper that removes the task from `task_list` in `finally`. Completed, failed, and cancelled tasks are released immediately, including the final task when no later task is created, while running tasks stay tracked and task results and errors keep their existing behavior.

The compatibility pass covered in-tree callers, the MicroPythonOS docs, and organization-wide GitHub code search. Callers use both fire-and-forget tasks and retained task handles (weather refresh and Nostr lifecycle); those patterns remain compatible.

## Test plan

- `make lint`
- `make syntax-tests`
- `./scripts/test_runner.py tests/test_task_manager.py tests/test_task_supervision.py`
- `./scripts/test_runner.py tests/test_osupdate.py tests/test_webserver.py tests/test_websocket.py`
- Ran two successive batches of 100 completed tasks with 10 KiB buffers: `task_list` returned to empty and free heap remained identical at 30,549,952 bytes after GC.

A complete `make tests` run could not finish locally: the clean desktop build fails compiling the bundled SDL against the installed macOS SDK, and the fallback binary rejects source files containing restored native/viper decorators. The focused task lifecycle, supervision, OS update, web server, and live websocket tests passed.

### DIFF
--- a/internal_filesystem/lib/mpos/task_manager.py
+++ b/internal_filesystem/lib/mpos/task_manager.py
@@ -49,8 +49,17 @@ class TaskManager:
         cls.disabled = True
 
     @classmethod
+    async def _run_task(cls, coroutine):
+        task = asyncio.current_task()
+        try:
+            return await coroutine
+        finally:
+            if task in cls.task_list:
+                cls.task_list.remove(task)
+
+    @classmethod
     def create_task(cls, coroutine):
-        task = asyncio.create_task(coroutine)
+        task = asyncio.create_task(cls._run_task(coroutine))
         cls.task_list.append(task)
         return task
 
--- a/tests/test_task_manager.py
+++ b/tests/test_task_manager.py
@@ -1,3 +1,4 @@
+import asyncio
 import unittest
 from mpos.task_manager import TaskManager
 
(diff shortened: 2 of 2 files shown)


ThomasFarstrike:
Closing in favor of #266 — they are identical commits (same OID) fixing the same issue. #265 was accidentally created as draft.

PR #266, opened 2026-08-21
Fix TaskManager finished tasks memory leak

### LINKED ISSUE #261: TaskManager retains completed tasks indefinitely (memory leak)

fdb:
This PR fixes a memory leak in `TaskManager`. `TaskManager` previously retained every task created through `create_task()`, which also kept completed coroutine frames and their captured objects alive indefinitely.

We fix this by removing managed tasks from `TaskManager.task_list` as soon as they finish (using try/finally) so completed coroutine frames and captured objects can be garbage collected. Preserve task results, exceptions, cancellation, and running-task tracking.

We added multiple tests in `test_task_manager.py` to check the behavior and catch regressions.

This fixes #261.

### DIFF
--- a/internal_filesystem/lib/mpos/task_manager.py
+++ b/internal_filesystem/lib/mpos/task_manager.py
@@ -49,10 +49,19 @@ class TaskManager:
         cls.disabled = True
 
     @classmethod
+    async def _run_task(cls, coroutine):
+        task = asyncio.current_task()
+        try:
+            return await coroutine
+        finally:
+            if task in cls.task_list:
+                cls.task_list.remove(task)
+
+    @classmethod
     def create_task(cls, coroutine):
         if cls.disabled:
             return None
-        task = asyncio.create_task(coroutine)
+        task = asyncio.create_task(cls._run_task(coroutine))
         cls.task_list.append(task)
         return task
 
--- a/tests/test_task_manager.py
+++ b/tests/test_task_manager.py
@@ -1,3 +1,4 @@
+import asyncio
 import unittest
 from mpos.task_manager import TaskManager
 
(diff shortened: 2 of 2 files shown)


ThomasFarstrike:
Another step in the right direction!

PR #267, opened 2026-08-21
Show Battery app: reuse LVGL drawing objects

fdb:
The "Show Battery" app is a simple app that is often used as an example app for other apps. However, testing found out it has quite a lot of memory churn.

This is caused by making new LVGL line descriptors for every single line segment. This results in hundreds of allocations for each frame.

In this PR we reuse one line descriptor and two point objects for all the graph segments. Additionally, we batch each graph refresh into a single canvas layer lifecycle. There is also an ineffective canvas child cleanup call that we removed.

Since this is an example app, showing how to re-use a descriptor might show performance best practices that can improve other apps well.


ThomasFarstrike:
Excellent!

Issue #268, opened 2026-08-21
main no longer fits the 3.5 MiB OTA partitions: esp32s3 app image is ~44 KB over


bitcoin3us:
Building current `main` (ca199f6d) for esp32s3 fails the final size check:

```
Error: All app partitions are too small for binary micropython.bin size 0x38b120:
- Part 'ota_0' 0/16 @ 0x20000 size 0x380000 (overflow 0xb120)
- Part 'ota_1' 0/17 @ 0x3a0000 size 0x380000 (overflow 0xb120)
```

That's an app image of **3,715,360 bytes** against the 3,670,016-byte (3.5 MiB) slots set in `build_mpos.sh` for the fri3d devices — **45,344 bytes over**, so the next release currently can't ship for those partitions.

Data points:

- The pinned lvgl_micropython submodule (c3bbcfd3) **already includes** the compressed fonts from MicroPythonOS/lvgl_micropython#8 (−19 KB) — without them this would be ~64 KB over.
- The 0.17.3 release `.ota` was 3,654,784 bytes (15 KB of headroom), so `main` has grown by **~59 KB since 0.17.3** (roughly: QR deep links + FontManager rework + unihiker k10 board support + camera hardening + everything else merged this week).
- Repro: `./scripts/build_mpos.sh esp32s3` on current main (macOS host, idf 5.5).

Happy to help hunt for savings — an obvious first step would be a small CI check that fails PRs when the esp32s3 image exceeds the partition budget, so this gets caught at merge time instead of at release time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ThomasFarstrike:
Hmmm, on my end, it's still less than 3.5 MiB:

```
ls -al lvgl_micropython/lib/micropython/ports/esp32/build-ESP32_GENERIC_S3-SPIRAM_OCT/micropython.bin 

-rw-r--r-- 1 user user 3646368 Aug 21 14:25 lvgl_micropython/lib/micropython/ports/esp32/build-ESP32_GENERIC_S3-SPIRAM_OCT/micropython.bin
```

= 3.477 MiB = 23 KiB to spare... I didn't do a totally clean build from scatch but the latest from the build server (Linux CI server, clean build) for esp32s3 gets 3650720 bytes = 3.48 MiB = 18 KiB to spare.

It would indeed be surprising to see ~59 KiB geting added just from a few small code changes.

But maybe this is a difference between building on MacOS and Linux?

[...]

PR #270, opened 2026-08-21
builtin archive: strip dev junk and enable compression

bitcoin3us:
For #268 — which turned out to be a false alarm caused by my own environment (see [the correction](https://github.com/MicroPythonOS/MicroPythonOS/issues/268#issuecomment-5373414431)): a gitignored stray `__pycache__/appstore.cpython-314.pyc` (64,921 bytes of CPython bytecode from a desktop `py_compile` syntax check) was being swept into my local firmware images by `compile_dir.sh`'s `cp -RL`. Clean builds were fine all along. This PR turns that discovery into two fixes:

## 1. Junk-proofing the builtin archive (`compile_dir.sh`)

Strip `__pycache__`/`*.pyc`, `.DS_Store`, and editor backup files from the staging copy before archiving. These are gitignored — invisible in `git status` — so any contributor can silently ship them into firmware today; my 64 KiB `.pyc` was exactly that, and it manufactured a phantom release blocker. With this change the class of bug is gone.

## 2. Enable `--compress --wbits 14` on the freezefs mount

The variant already prepared (commented out) in `freezefs_mount_builtin.sh`:

| | bytes |
|---|---|
| clean archive, uncompressed | 233,379 |
| clean archive, compressed | 193,892 |
| **saved** | **39,487** |

Consistent with @ThomasFarstrike's own measurement (~36 KiB) in the issue.

## Resulting image (esp32s3, clean tree, both fixes)

| | bytes | headroom vs 3,670,016 slot |
|---|---|---|
| before (clean) | 3,650,720 (CI) | ~18.8 KiB |
| **after** | **3,610,704** | **~58 KiB** |

## Runtime cost of compression

Per freezefs's docs, binary-mode reads are streamed (no whole-file RAM cost); only text-mode opens decompress fully. Everything that reads `/builtin` at runtime is binary-mode (`.mpy` imports, icons, gzipped HTML) except ~0.5 KB `MANIFEST.JSON` reads. Functionally verified on the macOS desktop build: AppStore + camera graphical suites green (23 tests) and `test_graphical_launch_all_apps` cold-launches every builtin app from the compressed archive (only the by-design `errortest` fails). Offer stands to flash this build to my ESP32-S3 badge after merge and report absolute app-launch timings.

Follow-up suggestion from the saga, unchanged: an esp32s3 build job in CI — the partition size check is built into the build, so a CI job would catch both real growth and broken local environments at PR time.


bitcoin3us:
On-device validation of this PR, from flashing the resulting build (3,610,704 bytes, both OTA slots) to an ESP32-S3 badge:

- The **compressed /builtin archive mounts and serves correctly on hardware**: the OS boots to the launcher, builtin apps (Settings, AppStore, Camera) launch and run, and MANIFEST.JSON parsing / icon loading all work — i.e. every runtime consumer of the mounted archive exercised on-device, not just desktop.
- Flashing used exactly the "advanced offline update" procedure from the new docs (write the app image to `0x20000`/`0x3A0000`; partition table read back from the device first and verified to match). The data partition was untouched throughout.
- Related datum for the compression-cost question (different codepath — font glyph RLE, not freezefs — but same theme): an on-device A/B of compressed vs uncompressed montserrat_12 scrolling measured ~+8.8% render cost in a worst-case full-redraw benchmark, details in MicroPythonOS/lvgl_micropython#8. For *this* PR's archive compression the hot path is `.mpy` import at app launch, which is a one-time streamed read per launch rather than per-frame work; builtin app launches feel unchanged on the badge. If you want a timed before/after of app cold-launch, say the word and I'll measure it the same file-based way.

scripts/freezefs_mount_builtin.sh
```diff
@@ -36,8 +36,11 @@ if [ $result -ne 0 ]; then
 fi
 
 pushd "$mydir"/../freezeFS/
-python3 -m freezefs --target /builtin --on-import mount "$tempdir" freezefs_mount_builtin.py
-#python3 -m freezefs --target /builtin --on-import mount --compress --wbits 14 "$tempdir" freezefs_mount_builtin.py
+# --compress saves ~78 KB of firmware (issue #268). Safe for runtime use:
```
ThomasFarstrike: Doesn't it save ~39 KiB of firmware?

ThomasFarstrike (changes requested):

scripts/freezefs_mount_builtin.sh
bitcoin3us: Right you are — that comment predated the __pycache__ discovery; the commit message and PR text got corrected but this line didn't. Fixed in 11abbdf5 (~39 KiB).

PR #277, opened 2026-08-29
test tooling: fail loudly when unittest assertions are stripped

bitcoin3us:
## The problem

MicroPython's `unittest` implements every assert method with a bare `assert` statement, which `mpy-cross -O3` strips. On a desktop binary with the lib frozen at `-O3` (prod-style builds), boot imports `unittest` from the frozen lib before the test runner can prepend `lib/` to `sys.path`. From then on, **every assertion in every test is a silent no-op**: `assertTrue(False)` passes, and whole suites report vacuously green.

This was discovered while validating #276: its three new regression tests reported `ok` from `scripts/test_runner.py` on a frozen macOS binary both before and after the library fix, while a direct import-based run against the filesystem lib showed the true FAIL/PASS transition. Exceptions still surface (a `raise` inside a test shows as ERROR), which makes the false green especially convincing — the suite *looks* like it is executing and judging tests.

CI builds fresh dev binaries and is presumably unaffected; anyone running tests locally against a prod-style build gets meaningless results with no warning.

## The fix

Add a self-check right after `import unittest` in both generated harnesses (the import-based process runner and the paste-mode device runner in `scripts/mpos_controller.py`): verify `unittest.TestCase().assertTrue(False)` actually raises `AssertionError`. When it does not, print a clear diagnostic:

```
MPOS_TEST_RUNNER FATAL: unittest assertions are no-ops: unittest was imported from a build with assert statements stripped (mpy-cross -O3, e.g. the frozen lib of a prod firmware), so test results would be vacuously green. Use a dev build, or ensure a non-optimized lib/ is on sys.path before unittest is first imported.
```

and report `TEST WAS A FAILURE` instead of running the suite. (A `raise` was deliberately avoided — an uncaught exception in the pasted harness breaks the exec sentinel protocol and hangs the controller until timeout.)

## Tests

New `assertguard` section in `tests/cpython_mpos_controller.py`: a run of a test file containing one failing assertion must never be reported as a success. This is an invariant across build types:
- on a stripped build it passes via the new guard (verified on a frozen macOS binary: fails before this change with a vacuous green, passes after),
- on a dev build it passes because the assertion genuinely fails the run.

Existing `basic` and `helpers` sections still pass (26/26).

## Merge checklist

- CHANGELOG.md: entry added under Testing in "Future release".
- Tooling-only change; no manifests, docs pages, or boards affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,8 @@ OS:
 - sdl_keyboard: CTRL-SHIFT-S (CMD-SHIFT-S on macOS) saves a timestamped BMP screenshot in the current directory, at the screen's own pixel size instead of the scaled SDL window
 
 Testing:
-- test_runner/mpos_controller: self-check that unittest assertions actually raise before running a suite, and fail the run with a clear diagnostic when unittest resolves to a build with assert statements stripped (mpy-cross -O3, e.g. the frozen lib of a prod firmware) — previously every assertion was a silent no-op and whole suites reported vacuously green
+- test_runner/mpos_controller: self-check that unittest assertions actually raise before running a suite; when boot cached a stripped copy (mpy-cross -O3, e.g. a frozen lib imported before lib/ is on sys.path), repair it by grafting the working assert methods from the filesystem lib/unittest, and only fail the run with a clear diagnostic when no working unittest is available — previously every assertion was a silent no-op and whole suites reported vacuously green
+- mpos_controller: write paste-mode payloads in chunks, draining the echoed input between chunks — a large payload written in one call deadlocked once payload plus echo filled the PTY buffers, hanging exec() until timeout
 - tests: eliminate all direct lv.task_handler() calls from the topmenu drawer test; wait_ms now uses pure time.sleep() instead of a tight lv.task_handler() polling loop, preventing an unrecoverable hang when SDL event polling blocks on macOS ARM CI after many process cycles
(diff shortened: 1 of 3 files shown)


bitcoin3us:
Pushed a follow-up commit that upgrades the guard from fail-loudly to self-heal-when-possible:

1. **Repair instead of abort when the filesystem lib/unittest is intact.** The common broken case is boot caching a stripped frozen `unittest` before `lib/` reaches `sys.path`. The harness now imports the filesystem `lib/unittest` fresh and grafts its working assert methods onto the cached module's `TestCase` — which also repairs classes that subclassed it at boot (e.g. `GraphicalTestCase`). The loud FATAL abort remains for builds with no working `lib/unittest` at all.

2. **Fixed a PTY echo deadlock the larger self-check exposed**: `exec()`/`exec_streaming()` wrote the whole paste-mode payload in one call, but aiorepl echoes every pasted byte, so once payload + echo filled the PTY buffers the write blocked forever. Payloads are now written in chunks with the echo drained between chunks.

Verified on a macOS build whose boot caches a stripped frozen unittest: a failing assertion now produces a real AssertionError traceback from `lib/unittest` and `TEST WAS A FAILURE`; `test_runner.py tests/test_shared_preferences.py` runs honestly (58 real tests, repaired-asserts notice printed); assertguard/basic/helpers controller sections pass 27/27.

PR #280, opened 2026-08-29
Add board support: Waveshare ESP32-S3-Touch-LCD-3.5

bitcoin3us:
Adds the 3.5" sibling of the supported ESP32-S3-Touch-LCD-2: [Waveshare ESP32-S3-Touch-LCD-3.5](https://www.waveshare.com/wiki/ESP32-S3-Touch-LCD-3.5) — 320×480 IPS (ST7796, SPI), FT6336 capacitive touch, QMI8658 IMU, PCF85063 RTC, AXP2101 PMU, ES8311 codec, TF slot, camera connector (OV5640/OV2640 supported, none included).

## What's in the change

- **`mpos/board/waveshare_esp32_s3_touch_lcd_3_5.py`** — modeled closely on the LCD-2 board file. The key hardware difference: **LCD reset and chip-select are on a PCA9554 I2C IO expander** (EXIO1/EXIO2 @0x20), not GPIOs — the board file pulses reset and parks CS low via the expander before SPI display init (`cs=-1`; the panel is the only SPI device). Battery status comes from the AXP2101 (same pattern as lilygo_t_watch_s3_plus) instead of an ADC pin.
- **Vendored `drivers/display/st7796/`** from lvgl_micropython's api_drivers, with the same package glue as the existing st7789 (ft6x36 was already vendored).
- **New `drivers/io_expander/pca9554.py`** — minimal 8-bit sibling of the existing tca9555 driver.
- **Auto-detection** in `mpos/main.py`: PCA9554@0x20 **and** FT6336@0x38 on sda=8/scl=7 — a pair unique to this board among supported ones. Probed **last** in the S3 chain, with `restore_i2c` on miss, so existing boards see no behavior change.

## ✅ Verified on the physical board

Flashed to an actual ESP32-S3-Touch-LCD-3.5 and confirmed working **on first boot**:

- Display initializes via the expander reset/CS sequence: boot screen and launcher render correctly
- Orientation and colors correct (landscape, BGR + byte swap as configured)
- **Touch accurate** — taps land where the finger is, UI fully navigable
- Board auto-detected (it's the only path that brings up this panel)
- AXP2101 alive: `has_battery: True`, percentage `-1` (correct PMU answer for USB-only power, no LiPo attached)
- Build: `./scripts/build_mpos.sh esp32s3` passes, size check reports image 3,618,496 bytes / 51,520 headroom (~7.8 KB cost for the board support)

Still untested for lack of accessories rather than doubt: LiPo battery charging/percentage with an actual battery, TF card, camera on the connector, IMU mounted-orientation sign (initialized and detected, orientation needs a tilt test). The board file is commented at each of those points.

## Pin sources

GPIO map cross-checked between the Waveshare wiki/schematic ([PDF](https://files.waveshare.com/wiki/ESP32-S3-Touch-LCD-3.5/ESP32-S3-Touch-LCD-3.5-Schematic.pdf)) and two independent community configurations — and now validated by the hardware itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Future release (next version)
 =====
 
 Board Support:
+- Add Waveshare ESP32-S3-Touch-LCD-3.5 (3.5" 320x480 ST7796 over SPI with LCD reset/CS on a PCA9554 IO expander, FT6336 touch, QMI8658 IMU, AXP2101 battery management); vendors the st7796 display driver and adds a PCA9554 expander driver
 - UNIHIKER K10: improve two-button navigation, correct camera preview orientation, and restore button input after camera operations
 - unix/macOS/web: compiler fallback to bytecode on architectures without a native emitter (e.g. macOS on Apple Silicon) instead of runtime-loaded apps using @micropython.native/@micropython.viper failing to import with SyntaxError 'invalid micropython decorator'
 
--- a/internal_filesystem/lib/drivers/display/st7796/__init__.py
+++ b/internal_filesystem/lib/drivers/display/st7796/__init__.py
@@ -1,0 +1,26 @@
+import sys
+from . import st7796
+from . import _st7796_init
+
+# Register _st7796_init in sys.modules so __import__('_st7796_init') can find it
+# This is needed because display_driver_framework.py uses __import__('_st7796_init')
+# expecting a top-level module, but _st7796_init is in the st7796 package subdirectory
+sys.modules['_st7796_init'] = _st7796_init
+
+# Explicitly define __all__ and re-export public symbols from st7796 module
+__all__ = [
+    'ST7796',
+    'STATE_HIGH',
+    'STATE_LOW',
+    'STATE_PWM',
+    'BYTE_ORDER_RGB',
+    'BYTE_ORDER_BGR',
+]
+
+# Re-export the public symbols
+ST7796 = st7796.ST7796
+STATE_HIGH = st7796.STATE_HIGH
+STATE_LOW = st7796.STATE_LOW
+STATE_PWM = st7796.STATE_PWM
+BYTE_ORDER_RGB = st7796.BYTE_ORDER_RGB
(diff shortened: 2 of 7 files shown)


bitcoin3us:
Hardware verification done — flashed the branch build to a physical ESP32-S3-Touch-LCD-3.5: display, orientation, colors, and touch all correct on first boot, board auto-detected, AXP2101 responding (details folded into the PR description). Only battery-attached charging, TF card, camera, and IMU orientation-sign remain untested for lack of accessories.

bitcoin3us:
Pushed one more commit: the vendored st7796 init was missing INVON (0x21), so the whole panel rendered in negative — an inverted UI passes for a dark theme, which is why it survived initial bring-up testing. It only became obvious once a camera was attached and the preview looked like an x-ray (an inverted photo). The st7789 driver already sends INVON for the LCD-2's IPS panel; this brings st7796 in line. Verified on hardware: with this init the launcher shows its true light theme and the camera preview (upcoming PR) renders natural colors.

bitcoin3us:
Evidence for the INVON commit, captured during debugging:

**On-device LVGL snapshot taken while the physical panel showed this exact scene in negative** — launcher and RGB565 color-bar test image are color-correct in the render buffer, isolating the fault to the panel init (a photo of the panel at the same moment showed cyan/magenta/yellow bars, i.e. exact bitwise inversion):

LVGL snapshot, correct colors

**Live camera preview frame tapped out of the running Camera app** (240x240 RGB565 LE, OV5640) — natural colors, proving the sensor/capture path was never the problem:

camera frame, natural colors

bitcoin3us:
One more hardware-verified update: display orientation changed from `_90` to `_270` — with the USB cable to the left and the BOOT/RESET buttons at the bottom (the board's natural desk position) the UI was upside down. Verified on the device: upright UI, touch aligned. One quirk worth knowing: rotation only takes effect at display init — setting it at runtime updates driver state but not what the panel shows, so this has to be right in the board file.

PR #281, opened 2026-08-30
Waveshare ESP32-S3-Touch-LCD-3.5: ES8311 audio support

bitcoin3us:
**Stacked on #280** (the board support PR) — merge that first; only the last commit is new here. As requested, this lands the board's native audio once the board itself is approved.

## What

Wires the existing `drivers/codec/es8311.py` into the new board file: **Speaker** output and **Microphone** input registered with AudioManager over I2S, with pop suppression via the codec's DAC soft-mute in the `on_open`/`on_close` hooks (the freenove ES8311 pattern, minus the amp-enable pin — Waveshare's examples show none for this board).

- **I2S pins from Waveshare's own demo code** (`01_audio_out` / `04_es8311_example` in the wiki ZIP): MCLK=12, BCLK=13, LRCK=15, playback=16, mic=14; codec at I2C `0x18` on the shared sda=8/scl=7 bus.
- The codec driver expects `machine.I2C`-style calls; a tiny adapter wraps the board's existing `lcd_bus` I2C so no second bus is opened on the same pins.
- **DAC volume defaults to 76%, tuned by ear on the physical board**: the driver's 85% default audibly distorts through the speaker connector, 70% is clean but quiet, 76% was the loudest clean setting in an A/B/x listening session.

## Verified on hardware

- ✅ WAV playback through an attached speaker on a physical ESP32-S3-Touch-LCD-3.5 — clean audio at the shipped default volume
- ✅ Codec initializes on the shared I2C bus alongside the PCA9554/FT6336/AXP2101/QMI8658
- ⚠️ Microphone input is registered with correct pins but untested (nothing attached to monitor it with)

## Observed while testing (separate issue, not this board)

Chaining playbacks with short (~400 ms) gaps can silently wedge the I2S pipeline — `stream_wav` raises `ESP_ERR_INVALID_ARG` and subsequent plays are silent until reboot. Looks like an AudioManager/stream_wav reopen race; single plays are unaffected. Happy to file it separately with the repro.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Future release (next version)
 =====
 
 Board Support:
+- Add Waveshare ESP32-S3-Touch-LCD-3.5 (3.5" 320x480 ST7796 over SPI with LCD reset/CS on a PCA9554 IO expander, FT6336 touch, QMI8658 IMU, AXP2101 battery management); vendors the st7796 display driver and adds a PCA9554 expander driver
 - UNIHIKER K10: improve two-button navigation, correct camera preview orientation, and restore button input after camera operations
 - unix/macOS/web: compiler fallback to bytecode on architectures without a native emitter (e.g. macOS on Apple Silicon) instead of runtime-loaded apps using @micropython.native/@micropython.viper failing to import with SyntaxError 'invalid micropython decorator'
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Future release (next version)
 =====
 
 Board Support:
+- Waveshare ESP32-S3-Touch-LCD-3.5: ES8311 audio support (speaker output + microphone input over I2S, DAC volume default tuned by ear on hardware)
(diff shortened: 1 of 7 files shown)


bitcoin3us:
Follow-up from further audio testing on this board (parallel session working on the ClipTV app, findings shared):

- The playback-duration overshoot it found in WAVStream (fixed sleep of a full 32 KiB ibuf after the last write instead of actual queued-duration) was diagnosed and fixed in #282, **verified end-to-end through this PR's Speaker/ES8311 output on the physical 3.5" board** — e.g. a 22.05 kHz stereo clip went from 4185 ms to 2219 ms wall clock. Good cross-validation that the Output registration here behaves.
- The raw I2S path on these pins (sck=13, ws=15, sd=16) drains within ~1% of nominal for mono and stereo at 8/22.05 kHz — clocking is sound.
- The "slight distortion" noted in this PR's listening test has a new best suspect after that work: **MCLK generated via LEDC PWM at 256×fs has ~1% frequency granularity error** at these frequencies — i.e. codec-clock jitter, not the write path and not this board file. Worth a look someday for all PWM-MCLK boards (freenove too); not a blocker here.

Issue #287, opened 2026-08-31
NWC: NIP-47 error replies are silently dropped (app hangs on "Connecting..."); watchdog and reconnect compound it


bitcoin3us:
While debugging a Lightning Piggy device that sat on "Connecting to nwc backend..." forever, I traced the NWC receive path end to end (instrumented desktop 0.16.1 run, plus relay-side queries against relay.coinos.io from a separate client). The transport is fine; three real bugs surfaced in `nostr_service.py` / the NWC flow. Evidence and mechanisms below; a PR for bug 1 is attached separately.

## Bug 1: NIP-47 error replies are silently discarded
`NostrManager._process_nwc_event()` reads `response.get("result")` and `response.get("notification")` but never `response.get("error")`. When the wallet service answers with an error, nothing happens.

Observed concretely: a coinos NWC connection that had been retired server-side. Every `get_balance` / `list_transactions` request got a prompt reply:

```
{"result_type":"get_balance","error":{"code":"UNAUTHORIZED","message":"This NWC connection has been retired; create a new one"}}
```

The reply is received, decrypted... and dropped. The app shows its connecting state forever, indistinguishable from a dead relay. This cost a full day of debugging in the wrong layers (websocket, TaskManager, firmware A/B) because the failure looks exactly like a transport hang.

Fix (see PR): forward errors to the existing `_error_cb`, dedupe identical repeats (the same error otherwise re-fires every poll), and reset the silence watchdog since an error reply is wallet-service activity. Lightning Piggy already renders `error_cb` text on the payments label, so the message becomes user-visible with no app-side UI work (the app does need to wire `mgr.set_error_callback`, which `nwc_wallet.py` currently never calls; LightningPiggyApp PR does that).

## Bug 2: silence watchdog ignores replies it cannot use
`_polls_since_last_event` is only reset on a successful balance/transactions result. With the wallet answering promptly but unusably (the UNAUTHORIZED case above), the watchdog concluded "silent for 3 polls" and reconnected the relay in a loop:

```
WARNING:nostr_service:NostrManager: watchdog reconnecting relay (silent for 3 polls)
```

Reconnecting cannot fix an unauthorized connection, and it tears down a socket that is demonstrably working. The PR resets the counter on error replies; a broader fix could count any received event on the NWC subscription.

## Bug 3: `_reconnect_relay()` creates a message-pool split brain
The watchdog reconnect does:

```python
[...]

PR #288, opened 2026-08-31
Nostr: surface NIP-47 error replies instead of silently dropping them

bitcoin3us:
Fixes bug 1 of #287 (and the error-reply case of bug 2).

## Problem
`NostrManager._process_nwc_event()` reads `result` and `notification` but never `error`. A NIP-47 error reply, e.g.

```
{"result_type":"get_balance","error":{"code":"UNAUTHORIZED","message":"This NWC connection has been retired; create a new one"}}
```

was decrypted and then discarded. The app sat on its connecting state forever, indistinguishable from a dead relay, while the silence watchdog (which only counts successful results) pointlessly reconnected a relay that was answering within ~2 seconds. Full evidence chain in #287.

## Fix
On an `error` reply:
- forward a readable message (`NWC get_balance: This NWC connection has been retired; create a new one (UNAUTHORIZED)`) to the existing `_error_cb`
- dedupe identical repeats: the same error re-fires every poll, and re-painting the same message each minute is noise; a successful result re-arms forwarding
- reset `_polls_since_last_event`: an error reply IS wallet-service activity, and reconnecting cannot fix an unauthorized connection

Lightning Piggy already renders `error_cb` text on the payments label (the same label that shows "Connecting to nwc backend..."), so the message becomes user-visible in place. Note: `nwc_wallet.py` never calls `mgr.set_error_callback`, so a companion LightningPiggyApp PR wires that up and syncs this file per the usual convention.

## Testing
- `tests/test_nostr_nwc_errors.py` (new): error reaches the callback, watchdog reset, repeat-dedupe, re-arm after success, and no-callback safety. All 5 pass via `python3 scripts/test_runner.py tests/test_nostr_nwc_errors.py` on the desktop build.
- Manually reproduced the original failure and the fix against the real retired coinos connection on the macOS desktop build.

## Merge checklist
- CHANGELOG: ✅ Future release → Frameworks
- Unit tests: ✅ added and passing
- App version / MANIFEST, MAINTAINERS, settings migration, docs: n/a (framework behavior fix)
- Bugs 2 (general watchdog accounting) and 3 (reconnect MessagePool split-brain) of #287 are deliberately NOT addressed here to keep this reviewable; happy to follow up.


bitcoin3us:
Amended after end-to-end verification caught a dedupe flaw: `get_balance` and `list_transactions` errors alternate every poll, so a same-as-last-message check on the full display string forwarded every reply. The dedupe key is now the error itself (`(code, message)`), independent of `result_type`, with a new unit test covering the alternation case (6 tests total, all passing).

PR #290, opened 2026-09-01
deeplink: fold query keys to lowercase; accept www.badgehub.eu

bitcoin3us:
Small follow-up on top of 9e72e7f7 (the BadgeHub switch), which supersedes the original version of this PR. One commit:

- **Query parameter keys are case-folded** like the rest of the link: the module contract says QR alphanumeric-mode uppercasing is handled everywhere, but `?V=1.2` silently dropped the version parameter. Test added.
- **`www.badgehub.eu`** accepted alongside `badgehub.eu`, and reserved from third-party `urlPattern` handlers the same way.

(The optional legacy `apps.micropythonos.com` recognition was dropped per review — the feature was never officially released, so no such QR codes exist in the wild.)

Tests: `test_deeplink.py` plus the scanqr and camera-chip graphical suites pass on the desktop build.

### DIFF
--- a/internal_filesystem/lib/mpos/content/deeplink.py
+++ b/internal_filesystem/lib/mpos/content/deeplink.py
@@ -34,7 +34,7 @@ MAX_URL_LENGTH = 512
 MAX_APP_ID_LENGTH = 64
 MAX_PARAM_LENGTH = 64
 
-STORE_HOSTS = ("badgehub.eu",)
+STORE_HOSTS = ("badgehub.eu", "www.badgehub.eu")
 STORE_SCHEMES = ("micropythonos", "mpos")
 APPSTORE_FULLNAME = "com.micropythonos.appstore"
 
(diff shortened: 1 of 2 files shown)


ThomasFarstrike:
Thank you for all the pull requests!

I merged all of them *except* this one because I already took a stab at implementing it when I removed apps.MicroPythonOS.com support from the AppStore.

So this one has a few merge requests, some trivial like docstrings, some I'm not sure about... I might have missed some things myself which you did think of...

Therefore, would you mind having your clanker resolve the conflicts, or close the pull request if there's nothing more to do? Thanks again!

bitcoin3us:
Thanks for the merges! Resolved: rebuilt this on top of your BadgeHub switch instead of fighting the conflicts — your implementation covers the substance, so what's left is two small commits (details in the updated description). Commit 1 is the query-key case-fold bug fix plus `www.` acceptance; commit 2 keeps already-printed `apps.micropythonos.com/app/<id>` QR codes scannable on-device (recognition only, no fetch) and is deliberately separate so you can drop it if you'd rather retire that host completely. Both fine either way.

ThomasFarstrike:
Awesome! I'm not sure there's any use to supporting the old apps.MicroPythonOS.com links - I don't think anyone uses them because this feature hasn't been officially released so...?

bitcoin3us:
Agreed — dropped the legacy-host commit; the PR is now just the query-key case-fold fix plus `www.badgehub.eu` (single commit, tests green).

PR #298, opened 2026-09-15
InputActivity: optional selected_callback fired on every pick, before Save

bitcoin3us:
Adds an optional `selected_callback` key to the setting dict handled by `InputActivity`.

## Behaviour
- Called as `selected_callback(value)` on every radio pick, before Save: on a new selection, and again when the user re-taps the already-selected option (a preview hook wants to fire again). Also on dropdown value changes.
- Not fired when `allow_deselect` un-checks the active option (nothing is selected then).
- Exceptions inside the callback are logged and swallowed so a broken preview cannot break the input screen.
- `changed_callback` is untouched: it still fires only after Save when the stored value changed. Existing settings that do not set the key see no change.

## Why
Pickers that want to demonstrate a choice live (the motivating case: Lightning Piggy's payment sound picker, where tapping "Pig Oink" should play it) currently have to replace the stock radio screen with a custom Activity, because `changed_callback` only arrives after Save. With this key the stock screen does it in one line.

## Implementation notes
- The notify logic is a module-level helper (`_notify_selected(activity, index)`) rather than a method so `radio_event_handler` keeps working on the minimal fixtures used by `tests/test_graphical_setting_activity_radio.py`.
- Docs live in the separate MicroPythonOS/docs repository: bullets for the InputActivity and SettingActivity pages are in MicroPythonOS/docs#11 (same style as `allow_deselect`); merge together.

## Testing
- New `tests/test_graphical_input_activity_selected_callback.py` (6 cases: new selection, switching, re-tap fires again and stays checked, deselect does not fire, exception swallowed, fixture without the attribute). Passes on the rebuilt macOS desktop build.
- Existing `tests/test_graphical_setting_activity_radio.py` still passes (4/4).

## Merge checklist
- CHANGELOG: yes (Frameworks)
- Docs: yes, as MicroPythonOS/docs#11 (two framework pages)
- Tests: yes (new graphical test, existing one green)
- Version bump / MANIFEST / MAINTAINERS / migrations: n/a (additive framework key, default None)


ThomasFarstrike:
Ow yeah! This would also be great to have for setting the global notification sound in the Settings app!

bitcoin3us:
Good idea — added in 1 commit on this branch: the Settings "Notification" sound row now sets `selected_callback` to a new `NotificationManager.preview_sound(rtttl)`, so each option plays once on the buzzer as you tap it, before Save. `preview_sound` ignores the stored preference and the rate limiter; the "None" option or a board without a buzzer is a silent no-op. CHANGELOG entry and 3 unit tests included; `test_notification_manager.py` and the selected_callback graphical tests pass on a desktop build of this branch.

ThomasFarstrike:
Wow, props for also applying this new feature for the notification sound setting, adding a test, the whole shebang!

PR #300, opened 2026-09-16
uaiowebsocket: don't send a second pong on incoming pings

### LINKED ISSUE #299: uaiowebsocket: 'Failed to send pong' error on every relay ping (aiohttp already replies; ClientWebSocketResponse has no pong())
Seen on an ESP32-S3 running Lightning Piggy with an NWC relay connection, logged repeatedly on the serial console:

```
ERROR:uaiowebsocket:Failed to send pong: 'ClientWebSocketResponse' object has no attribute 'pong'
```

## What happens
`internal_filesystem/lib/uaiowebsocket.py` handles an incoming PING frame in `_connect_and_run()` by calling `await self.ws.pong(data)`. The bundled aiohttp port (`micropython-lib/python-ecosys/aiohttp/aiohttp_ws.py`) has no `pong()` (nor `ping()`) on `ClientWebSocketResponse`; its surface is `send_str`, `send_bytes`, `send_json`, `receive_*`, `close`. So every relay ping raises AttributeError, caught and logged at ERROR.

## Why it is noise rather than a dropped connection
aiohttp's `WebSocketClient.receive()` already answers pings itself: `_process_websocket_frame(PING, payload)` returns `(PONG, payload)` and `receive()` does `await self.send(data, send_opcode)` before returning. The PING frame is then still handed up to the caller (because its payload is non-empty), which is what triggers uaiowebsocket's PING branch. Net effect: the relay does get its pong, and then uaiowebsocket fails to send a second one and logs an error. Relays stay connected; the log is misleading.

Introduced with a548bb76 ("Nostr: fix incoming, add tests").

## Secondary finding: the module's own ping/pong timeout cannot work on this stack
- `_send_ping_async()` sends the keepalive as a BINARY data frame (`send_bytes(b"ping")`), not a PING control frame, so a server never replies with a PONG to it.
- aiohttp swallows incoming PONG frames (`_process_websocket_frame(PONG)` returns `(None, None)`, never returned to the caller), so uaiowebsocket's `msg.type == OPCODE_PONG` branch never runs and `last_pong_tm` never updates.
- Today `ping_timeout` defaults to `None`, so this path is dormant; if anyone enables it, every connection would be declared timed out.

## Suggested fix
1. In the PING branch, drop the redundant `await self.ws.pong(data)` (keep the `on_ping` callback), since aiohttp already replied. If an explicit reply is preferred, the working call on this port is the underlying client's `send()` with the PONG opcode: `await self.ws.ws.send(data, self.ws.ws.PONG)`.
2. If the keepalive is meant to be a real websocket ping, send it as a control frame the same way (`await self.ws.ws.send(payload, self.ws.ws.PING)`), and note that PONG replies will not surface to the caller unless aiohttp's `receive()` is changed to return them.

Happy to send a PR for (1), it is a two-line change; (2) depends on whether the keepalive/timeout feature is wanted.

bitcoin3us:
Fixes #299.

## Problem
On every relay PING the console logged:

```
ERROR:uaiowebsocket:Failed to send pong: 'ClientWebSocketResponse' object has no attribute 'pong'
```

The bundled aiohttp port already answers pings inside `WebSocketClient.receive()` (it maps PING to a PONG send before handing the frame up), and its `ClientWebSocketResponse` has no `pong()` at all. So the manual `await self.ws.pong(data)` in `_connect_and_run()` could never succeed; it raised and logged a misleading error each time. Connections were never affected (the relay had its pong).

## Fix
The PING branch now only runs the `on_ping` callback, through a small `_handle_ping()` helper whose docstring records why no reply is sent. Two lines of behaviour change, no API change.

Not touched here (noted in #299 as a secondary finding): the module's own keepalive sends "ping" as a binary data frame rather than a PING control frame, and aiohttp swallows real PONG frames, so `ping_timeout` cannot work on this stack; it is dormant (default `None`).

## Testing
- New `tests/test_uaiowebsocket_ping.py` (3 cases): pins that the bundled aiohttp has no `pong()`, checks `_handle_ping` still fires `on_ping` through the callback queue, and guards against the call returning. Loads the module from `lib/` so it tests the source without a rebuild. Passes.
- Observed on the ESP32-S3 board before the fix: the error on every relay ping while the relay stayed connected, consistent with the analysis.

## Merge checklist
- CHANGELOG: yes (Frameworks)
- Tests: yes
- Docs / version / MANIFEST / migrations: n/a


ThomasFarstrike:
Nice one!

May I ask what the urgency is, or the impact? To know how soon I should build a new release for this :-)

bitcoin3us:
No urgency — impact is cosmetic. Connections were never affected (the bundled aiohttp already answers the PING before the frame reaches uaiowebsocket, so the relay always had its pong); the only cost is an ERROR line on every relay ping, which on a Nostr/NWC-heavy app like Lightning Piggy means a steady stream of misleading errors on the serial console that buries real ones while debugging. Fine to ride along in the next regular release; no hotfix needed.

PR #302, opened 2026-09-16
SettingsActivity: render non-string setting values as text (fixes "app threw an exception" on settings pages)

bitcoin3us:
## Problem

Since 62e1a099 (`settings: dont_persist rows with default_value show (defaults to X)`) the settings list hands the stored pref value from `_row_value_text()` straight to `lv.label.set_text()`. Prefs are JSON, so a value that was written as a number, by an older app version or through numeric `ui_options` values, comes back as an `int`, and `set_text` raises:

```
TypeError: can't convert 'int' object to str implicitly
  File "mpos/ui/settings_activity.py", line 117, in onResume
```

The user sees the "App threw an exception. It might be glitchy. Maybe there's an update that fixes it?" dialog every time that app's settings page opens. Reproduced with ClipTV's `buttons_per_screen` (stored as `6`) on a Waveshare ESP32-S3-Touch-LCD-3.5 running current main.

## Fix

`_row_value_text()` now returns `str(...)` of the label-mapped value, so any stored type renders as text; the `(defaults to X)` paths were already safe through their f-strings. One line, plus four unit tests in `tests/test_settings_activity_label_mapping.py` (int value, int value mapped to a `ui_options` label, int default, strings unchanged) and a changelog entry.

## Verification

- `tests/test_settings_activity_label_mapping.py`: 27 tests pass on the desktop build.
- On the badge, opening ClipTV's settings after the fix is loaded shows the row as "6" with no dialog.

With thanks to the scientists and engineers who did the hard, unglamorous work that got us here.


ThomasFarstrike:
Nice catch! And great that you added a test for this because clearly the current tests missed it!

Question: is this an urgent fix for the conference? Or is it only showing up in ClipTV which can probably work around it until the next release...?

bitcoin3us:
Sorry for the slow reply — not urgent from our side. It's a regression since 62e1a099 that hits any app whose settings page lists a pref stored as a number; ClipTV's `buttons_per_screen` is where it reproduced, and ClipTV can carry a trivial workaround (store that value as a string) until the fix ships, so no hotfix release is needed on our account. Several builtin apps also `put_int` prefs, but only ones that expose such a key through a SettingsActivity page would hit it, and I'm not aware of one that does. Fine to ride the next regular release.

